Issue #2, opened 2024-06-13
Explore whether ProgPy is a good base


WardLT:
ProgPy is a code developed by NASA for fitting and deploying prognostic models of many different kinds of dynamical systems.

They already have a well-organized class structure for models of systems and state estimation algorithms, with several implementations of algorithms we will use in AutoSOH. 

Challenges:
- [ ] There is no differentiation between state parameters which represent the dynamic state of the system, and those which could capture long-term degradation. I'm exploring a solution in https://github.com/ROVI-org/auto-soh/tree/progpy-explore
- [ ] Class implementations use some practices that make modular codes difficult (e.g., overriding Python's serialization logic, relying heavily on class attributes). Starting to untangle them in https://github.com/ROVI-org/progpy/tree/allow-dynamic-parameters
- [x] Output function for prognosis models lacks the control signal as an input. Attempted a PR in nasa/progpy#142

WardLT:
Pull request #3 mimicks ProgPy, but with some changes to how the base classes are implemented to better match how we've been approaching SOH.

I'll hope to close the delta between their base classes and ours enough that I can drop ours, but am moving forward with #3 to better understand our requirements.

victorventuri:
I think we have determined this not to be the case, right? @WardLT Can we close this issue?

WardLT:
I'd still like to keep it open. I closed that issue because ProgPy suggested a nice workaround.

I'm still keeping a mind open that we may be able to merge with ProgPy or PyBOP in the future, even though I think we should continue here to get a better of idea of what we need.

PR #3, opened 2024-06-16
Progpy-inspired base class using Pydantic

WardLT:
An implementation of a ProgPy-like interface using Pydantic to define the the containers which store state, control, and outputs from a model.

See #2 for further discussion on this topic.

To do list:
- [x] Have the Observer return some diagnostics (e.g., `y_err`) for each step
- [x] Decide how we should vary control inputs between last and current step. (Deferred, see #4)
- [x] Document how to create a new State class
- [x] Document how to use the Observer
- [x] Give unique names to each parameter within a sequence of parameters
- [x] ~~Moving covariance to a property of the filter~~ Revisiting later
- [x] Adjust terminology
  - Observer to online estimator
  - Control to Inputs
  - Output to Measurement
- [x] Switch back to Py3.9
  

There are more to-dos in the code, but they would be better-suited for later PRs


WardLT:
@victorventuri , I'll be ready for your review soon. The points I know I need your input on include: 

- [ ] Terminology: Things like "control," "observer," "state," and "state of health" are things we can change later but now is probably a good idea to get them right now
- [ ] What kind of things should I return from the UKF implementation as diagnostics? Right now, it only returns the error between estimated and observed outputs
- [ ] Does the process for defining a new model make sense? 
- [ ] What about for defining a new "Observer?" Is there another easy one we should implement to sanity check?

Issue #7, opened 2024-06-21
Decide where the definition of which parameters are "state of health" should be


WardLT:
Right now, we list which parameters are being treated as state of health as part of the "SystemState" class.

However, many of our strategies for state estimation may rely on updating different parameters during different phases. Thus, it feels more likely that the list of which parameters to be updating should lie with the class that is using that information -- `OnlineEstimator`. 

If we do that, then the covariance matrix between each parameter should be an attribute of the OnlineEstimator as well because the State will not have knowledge of which parameters require covariances.

victorventuri:
I think that makes sense. I think we'll just need to make sure to provide an (optional?) argument to the several `to_numpy()` methods to list which `Field`s should be used when building the `np.ndarray` (which is something you partially have on your current version of `SystemState`)

WardLT:
Such a "get me only these variables" function would make a lot of sense for the `JointState`

victorventuri:
Should we close this? I think we reached a good place in the definition of `HealthVariable`

WardLT:
Yea, let's close

Issue #6, opened 2024-06-21
Make separate data models for "state" and "parameters"


WardLT:
Our current `SystemModel` requires users to annotate which attributes belong to different categories, such as "state" or "health" parameters. It would be cleaner to have those separated into two subclasses.

Current:

```python

class MySystemModel(SystemModel):
      attr_1
      attr_2
      state_params = ('attr_1',)
```

Proposed

```python

class State:
    attr_1

class Health:
    attr_2

class JointState():
    state = State()
    att_2 = Health()
```

victorventuri:
Agreed, will look into starting that process early next week!

WardLT:
Fixed by #11

PR #11, opened 2024-07-03
Refactoring the Health Variable Class

### LINKED ISSUE #10: New code on ecm_redefine is too complex
Class to hold on to HealthVariable objects has is too complex. How can we rewrite it such that the following **requirements** are met
- [ ] It contains an unspecified number of HealthVariable objects, determined by the end application (such as multiple RC components)
- [ ] It knows which HealthVariable objects can be updatable
- [ ] It can update specified objects from arrays, similarly to what HealthVariable does
- [ ] For a hypothetical end user, it can be treated as a HealthVariable object, seeing as these are conceptually equivalent

The **questions** that I have are:
1. Can we simply implement all of this functionality inside of HealthVariable, like on commit b1c53bc?
2. If not, can we have attributes be HealthVariables or lists/tuples of HealthVariables? This way, we could clean up a lot of the dynamic allocation code
3. (in the not-so-distant future, can we make sure we can have collections of collections of HealthVariables in order to implement efficient operations with numpy.ndarrays?)

Additional **comments**:
- Part of the reason why I did it dynamically was the following hypothetical situation: imagine that, in the future, we still want to use the ECM, but with some physics included, say, we want to incorporate loss of lithium inventory and loss of active material. We could simple '+=' those HealthVariables into the ECMASOH without needing to rewrite another whole class to deal with this. However, perhaps it would be best if these were all completely separate classes as well.  

@WardLT What other things am I missing from this?

WardLT:
Implements a hierarchical model for the parameters of different storage system models.

Fixes #10 by providing a simpler implementation of the features there (similar to Option 1 of the proposed) and #6 by creating a class which is only for model parameters. 

We seem to be drifting towards an solution to #7 where the updatable parameters are kept alongside the model parameters, which I'm OK with thus far. That may change once we introduce some kind of container for "Joint State" which combines the always-updatable model state with sometimes-updatable 

To do list:
- [x] Add a "set item by name" method
- [x] Raise errors if users try to update an field is not marked as updatable 
- [x] Express a HealthVariable for an ECM with variable numbers of RC elements
- [x] Get all updatable variables
- [x] Get total length of updatable parameters
- [x] Express different models for circuit elements whose properties change with SOC/temperature
- [x] ~~Add TODOs in the codebase indicating where to add a "batch dimension" for when a single HealthVariable instance describes a several parameterizations of the same physical system. Create an Issue describing what behavior we want~~ _We'll deal with this later_
- [x] ~~Add a new data model for system state and outputs~~ Punting for now, use the GeneralContainer from @victorventuri 's refactor
- [x] Either delete existing "Model" and "UKF" implementations then create an Issue to add them back in, or modify them (if we're successful in this implementation)
- [x] Disallow lists, only allow sequences to be tuples


victorventuri:
I am jotting down comments here as I read through the code, so maybe by the time I reach the end, some of them will have been resolved and I may forget to iterate back. 

1. Based on the `make_all_updatable` method, the current definition of `HealthVariable` does not allow for fields to just be lists (or iterables) of floats, for instance; everything inside it must be another `HealthVariable` it looks like 
2. Regarding the TODO on line 68 of `base.py`, the answer is positively "yes!": when assembling the necessary matrices for updating transient states, we need to know the values of parameters even if they are "immutable" 
3. I like the option of making everything updatable recursively (despite somewhat fearing it a bit as well)

PR #23, opened 2024-07-10
Base definitions for online estimators

victorventuri:
Here, we have base definitions for online estimators. The main idea is that the estimators should be fully numeric, and totally agnostic to the models being used (though they must know what numbers/`numpy.ndarray`s are inputs, hidden states, and outputs, of course).

 We have a `MultivariateRandomVariable` class, which is used to describe hidden states and measurements (maybe also inputs, but I don't know how to properly do that yet). We also establish a class to facilitate the interface between `CellModel`s and the estimators.

### DIFF
--- a/asoh/estimators/__init__.py
+++ b/asoh/estimators/__init__.py
@@ -1,0 +1,1 @@
+""" Collection of state estimators """
--- a/asoh/estimators/online/__init__.py
+++ b/asoh/estimators/online/__init__.py
@@ -1,0 +1,1 @@
+""" Collection of online estimators """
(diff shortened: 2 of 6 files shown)


WardLT:
Could you show me how one creates a `OutputMeasurement` class? I'm still unsure why it needs to exist separately from an `OutputQuantities` if both represent kind of the same thing.

victorventuri:
The main idea is that the `OutputQuantities` is a deterministic output specific to the `CellModel`, while the `OutputMeasurements` is a random variable (so it would, for example, include mean and variance in the case of a Gaussian). Effectively, the `OutputMeasurements` acts as an interface between the "physics"-based `OutputQuantities` and the statistics-based estimator.

WardLT:
Ok, that makes sense. It seems sensible enough to keep measurements and model outputs separate for now, but I'm going to keep an eye open for an opportunity or reason to reduce complexity by merging the two.

victorventuri:
Sounds good! I think the best place to re-evaluate the distinction might be when we do the minor revamp for optimizing batches of hidden states (like from sigma points)

PR #26, opened 2024-07-15
Initial implementation of UKF

WardLT:
This PR implements, tests, and demonstrates the general UKF. In our demonstrations, we show how to use it on a battery system with constant A-SOH, and also apply it to a general system described by Lorenz equations.


asoh/estimators/online/__init__.py
```diff
@@ -1 +1,136 @@
...
+    """
+    Base class to help represent a multivariate random variable
+    """
+
+    @property
+    def expected_value(self) -> np.ndarray:
+        """
+        Provides expected value of distribution
+        """
+        if hasattr(self, 'mean'):
```
WardLT, line 24: Should we just add a `mean` property to this class if it's expected. Something like 

```python

@property
@abstractmethod
def mean(self) -> np.ndarray:
    pass
```

tests/estimators/online/general/test_ukf.py
```diff
@@ -0,0 +1,212 @@
+import numpy as np
+from pytest import fixture
+
+from asoh.estimators.online import ModelFilterInterface, ControlVariables, OutputMeasurements
+from asoh.estimators.online.general.kalman.unscented import UnscentedKalmanFilter as UKF
+from asoh.estimators.online.general.kalman import KalmanHiddenState
+
+
+# Define Lorenz dynamics
+class Lorenz(ModelFilterInterface):
```
WardLT, line 10: A thought about this test case: do you think we can still use the Lorenz model if we start making the `ModelFilterInterface` more closely integrated with `CellModel`? Can we make a `CellModel` that implements the Lorenz equations?

asoh/estimators/online/__init__.py
victorventuri: @WardLT I think that's a good idea! Going to push a new version soon.

tests/estimators/online/general/test_ukf.py
victorventuri: I thought a bit about that, but I think it would be weird to have a `CellModel`, with `abstractmethod`s `update_transient_state` and `calculate_terminal_voltage`, that does not do anything related to a battery cell.

asoh/estimators/online/__init__.py
victorventuri: Actually, the reason I did not do that is that, for some other child classes, `mean` is a `pydantic` `Field`, and not a property, so we cannot force it to be an `abstractproperty` in my opinion. @WardLT I don't quite see a way around that...

tests/estimators/online/general/test_ukf.py
WardLT: Let's see where we need to change it when we get there. I really would like to prevent users from needing to implement the same loops for `ModelFilterInterface` subclasses continually.

asoh/estimators/online/__init__.py
WardLT: Defining `mean` to be a pydantic field in subclasses also won't be a problem, as it will still provide the same "property" interface. Unless I'm mistaken, let's go ahead and add the `mean` property.

victorventuri: Fixed by enforcing `get_mean` abstractmethod

PR #32, opened 2024-07-18
All ECM components initialized fully frozen

### LINKED ISSUE #31: ECM A-SOH non-independent updatables
The standard template provider for the ECM A-SOH class has issues in which, by marking one of the sub-elements of the RC component as updatable, it would set both components updatable. This is being caused by the fact that some of the ECM components are initialized with a default `updatable` field already populated, rather than being an empty set. We should adopt the standard that everything in a `HealthVariable` should be considerable as frozen unless the user says otherwise, and not the other way around.

victorventuri:
Previously, some of the ECM components (such as the RCComponent) were being initialized with some inner fields updatable by default. This can cause issues, as the base HealthVariable class assumes everything is frozen unless the user says otherwise. Therefore, it made sense to maintain the updatable field empty for all HealthVariable children.

Fixes #31

### DIFF
--- a/asoh/models/ecm/components.py
+++ b/asoh/models/ecm/components.py
@@ -12,7 +12,6 @@ class MaxTheoreticalCapacity(HealthVariable):
     """Defines maximum theoretical discharge capacity of a cell"""
     base_values: float = \
         Field(description='Maximum theoretical discharge capacity of a cell. Units: Amp-hour')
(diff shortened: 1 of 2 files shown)


WardLT:
Other than another override for `updatable` in `ecm/utils.py`, this looks good to me. I removed that one and will merge once tests pass

WardLT (approved):

PR #38, opened 2024-07-23
Reduce the number of classes

WardLT:
Now that we've got the features, I'm seeing what we can do to reduce the complexity of the software by reducing the number of classes used.

The main goal of this branch is preventing the need to create a new ModelFilterInterface for each new model and each new filter, which requires a few major changes.

While at it, I'm ...
- [x] Eliminating subclasses which do not add new functionality (e.g., TransientVector, ASOH)
- [x] Making the ECM class conform to the interface defined by the base class (i.e., no args/kwargs, same argument names)
- [x] Update the notebooks to use the new interface
- [x] Combined the `ModelFilterInterface` with the OnlineEstimator. The `ModelFilterInterface` was never used without the estimator, so combining ensured that we need only create one variable (less code, opportunities for filters and estimators getting discordant information).
- [x] Detailed tests to ensure that normalization works in the JointUKF
- [x] Updating the docstrings to make sure they match the new arguments
- [x] Use "inputs" not "input" ~~or "control"~~ (I'll remove "control" once we're more comfortable with how the estimators and models fit together)


WardLT:
I'm getting linear algebra problems when we normalize. That didn't seem to be present before, and I cannot yet find the issue.

@victorventuri , how worried are you about normalization leading to problems?

moirae/models/ecm/__init__.py
```diff
@@ -10,18 +10,31 @@
...
-                               transient_state: ECMTransientVector,
-                               asoh: ECMASOH,
-                               previous_input: ECMInput,
-                               current_behavior: Literal['constant', 'linear'] = 'constant'
-                               ) -> ECMTransientVector:
+    current_behavior: str
+    """How it is assumed the current varies between timesteps"""
+
+    def __init__(self, current_behavior: Literal['constant', 'linear'] = 'constant'):
+        self.current_behavior = current_behavior
```
victorventuri, line 29: I am opposed to this. I like the idea you had earlier of the `ECM` model being fully standalone, so that it always works in the same way given what is provided to it. If it's ok with you, I'd prefer just keeping the `current_behavior` as an argument to the main methods this class has, and with the `'constant'` default.

moirae/models/base.py
```diff
@@ -427,20 +427,6 @@ class OutputQuantities(GeneralContainer):
...
-    pass
-
-
-class AdvancedStateOfHealth(HealthVariable):
-    """
-    Stores A-SOH
-    """
-    pass
-
-
```
victorventuri, line 443: I think removing the `TransientVector` is good. Now, I think we should have a discussion on the A-SOH part. While the A-SOH is, as the name implies, a `HealthVariable` on its own, containing other `HealthVariables`, it is one additional thing on top of that: it is the **necessary and sufficient** set of `HealthVariable`s to fully reproduce an IVT(t) profile. So, I believe there are merits to keeping these classes distinct, even if there is no added functionality in the A-SOH. I can be convinced otherwise though, so let's talk about this when we meet later today.

moirae/estimators/online/utils.py
```diff
@@ -2,6 +2,8 @@
 import numpy as np
 
 
+# TODO (wardlt): Why do we have two
+
```
victorventuri, line 6: Can probably be merged in a single function, I just thought this was more streamlined: one function to find the nearest positive semi-definite matrix, and one for figuring out if we need to use the algorithm or not. But, tbh, I can't foresee a case where they would not be used in tandem (one calls the other afterall...), so, if you feel like they can be implemented together, you can go ahead on that

moirae/models/ecm/__init__.py
WardLT: I disagree that being able to change the current behavior is sufficient reason to deviate from the interface of CellModel. 

Two alternatives:
- If the current behavior is indeed something which can change between calls to the update and output function, should make it part of the control variables? 
- If it is something that all CellModels will use, should we make it as argument in the base CellModel class?

moirae/estimators/online/utils.py
WardLT: Sorry, didn't realize that comment made it in there. I wrote it while having a brain fart and not realizing "enforce" and "ensure" are different words.

Let's keep it as-is.

moirae/models/ecm/__init__.py
```diff
@@ -116,10 +129,11 @@ def update_transient_state(new_input: ECMInput,
                                   i_rc=iRC_kp1,
                                   hyst=hyst_kp1)
 
-    @staticmethod
-    def calculate_terminal_voltage(new_input: ECMInput,
-                                   transient_state: ECMTransientVector,
-                                   asoh: ECMASOH) -> ECMMeasurement:
+    def calculate_terminal_voltage(
+            self,
+            inputs: ECMInput,
```
victorventuri, line 134: Isn't `input` a protected (but over-writable) name?

WardLT: Yes, so we use "inputs" here, which is not.

moirae/models/ecm/__init__.py
```diff
@@ -10,18 +10,31 @@
...
+    current_behavior: str
+    """How it is assumed the current varies between timesteps"""
+
+    def __init__(self, current_behavior: Literal['constant', 'linear'] = 'constant'):
+        self.current_behavior = current_behavior
+
+    def update_transient_state(
+            self,
+            previous_input: ECMInput,
+            current_input: ECMInput,
```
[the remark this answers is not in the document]
WardLT: lol! I did not think of that. I'll rename accordingly.

moirae/models/ecm/simulator.py
```diff
@@ -7,6 +7,7 @@
                                ECMMeasurement)
 
 
+# TODO (wardlt): Make this not part
```
victorventuri, line 10: I assume there was more to this sentence? Or maybe it should be obvious and I'm just not getting it?

WardLT: No, I probably just got distracted mid-sentence 🤦🏼

tests/estimators/online/test_ukf.py
```diff
@@ -0,0 +1,229 @@
...
+    rho: float
+    beta: float
+    n: float
+
+
+class LorenzOutputs(OutputQuantities):
+    m1: float
+
+
+class LorenzModel(CellModel):
```
victorventuri, line 27: fellowCellModel2

tests/estimators/online/test_ukf.py
```diff
@@ -0,0 +1,229 @@
...
+
+class LorenzControl(InputQuantities):
+    sigma: float
+    rho: float
+    beta: float
+    n: float
+
+
+class LorenzOutputs(OutputQuantities):
+    m1: float
```
victorventuri, line 24: I'm confused, didn't I use to have two values being output?

tests/estimators/online/test_ukf.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def calculate_terminal_voltage(
+            self,
+            inputs: LorenzControl,
+            transient_state: LorenzState,
+            asoh: HealthVariable) -> OutputQuantities:
+        hidden_states = transient_state.to_numpy()
+
+        return LorenzOutputs(
+            terminal_voltage=np.sqrt(np.sum(hidden_states ** 2)),
```
victorventuri, line 57: Oh, I see...

tests/estimators/online/test_ukf.py
WardLT: One is the terminal_voltage 🤫

moirae/estimators/online/distributions.py
```diff
@@ -1,11 +1,32 @@
...
+
+    @abstractmethod
+    def get_mean(self) -> np.ndarray:
+        """
+        Provides mean (first moment) of distribution
+        """
+        pass
+
+
+class PointEstimate(MultivariateRandomDistribution):
```
victorventuri, line 23: An "estimate" and a "distribution" are different things. I think the name here should be something like `MultivariateDeltaDistribution` for consistency.

victorventuri:
> I'm getting linear algebra problems when we normalize. That didn't seem to be present before, and I cannot yet find the issue.
> 
> @victorventuri , how worried are you about normalization leading to problems?

That's pretty weird, since I added the normalization to avoid linear algebra problems... It is not critical (the original version didn't have it!), but it should help considerably with the different orders of magnitude observed... So, all that is to say is that this is pretty puzzling: it should work better with normalization!

moirae/models/ecm/__init__.py
WardLT: Per my recollection of discussion, we're keeping it this way for now

moirae/models/ecm/__init__.py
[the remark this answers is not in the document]
WardLT: Renamed it to "new_inputs"

moirae/estimators/online/distributions.py
WardLT: Good call

tests/estimators/online/test_ukf.py
WardLT: 😂

PR #56, opened 2024-08-13
Introduce conversion between CellModel and ModelWrapper

WardLT:
Adds a class which fulfills the ModelWrapper interface given a cell model and list of ASOH parameters to include in the hidden state, as required by the Joint Estimator.


moirae/estimators/online/utils/model.py
```diff
@@ -0,0 +1,153 @@
...
+        input_template: Example input values for the model
+        asoh_inputs: Names of the ASOH parameters to include as part of the hidden state
+    """
+
+    def __init__(self,
+                 model: CellModel,
+                 asoh: HealthVariable,
+                 transients: GeneralContainer,
+                 input_template: InputQuantities,
+                 asoh_inputs: Tuple[str]):
```
victorventuri, line 75: Would it be possible to make this optional, and, when not provided, take everything that is updatable?

moirae/estimators/online/utils/model.py
```diff
@@ -0,0 +1,153 @@
...
+        Args:
+            hidden_states: Hidden states as used by the estimator
+        Returns:
+            - ASOH with values from the hidden states
+            - Transients state from the hidden states
+        """
+
+        # Update any parameters for the transient state
+        my_transients = self.transients.model_copy(deep=True)
+        batch_transients = np.repeat(self.transients.to_numpy(), axis=0, repeats=hidden_states.shape[0])
```
victorventuri, line 115: I'm a bit confused here; why are we repeating the transient state stored in `self`, instead of just doing `my_transient.from_numpy(hidden_states[:, :self.num_transients])`?

moirae/estimators/online/utils/model.py
```diff
@@ -0,0 +1,153 @@
...
+        previous_inputs = self.input_template.model_copy(deep=True)
+        previous_inputs.from_numpy(previous_controls)
+        new_inputs = self.input_template.model_copy(deep=True)
+        new_inputs.from_numpy(new_controls)
+
+        my_asoh, my_transients = self.create_cell_model_inputs(hidden_states)
+
+        # Produce an updated estimate for the transient states, hold the ASOH parameters constant
+        output = hidden_states.copy()
+        new_transients = self.model.update_transient_state(previous_inputs, new_inputs=new_inputs,
```
victorventuri, line 137: Should we pass `previous_inputs` as a keyword argument?

moirae/estimators/online/utils/model.py
WardLT: Totally. I just left the defaults off since I didn't know what sensible values would be

WardLT: i.e., and opted for being super-specific

moirae/estimators/online/utils/model.py
WardLT: Good point. I'm betting it's unnecessary and something I did out of abundance of caution over batching

moirae/estimators/online/utils/model.py
WardLT: That makes sense to me

PR #63, opened 2024-08-15
Renaming estimator wrappers for consistency

victorventuri:
Renamed some classes for consistency with the "wrapper" language, as well as to avoid confusion between cell model wrappers and degradation model wrappers.


WardLT (approved):

WardLT (changes requested):
You'll need to update the tests too

PR #66, opened 2024-08-16
Fix ECM broadcasting issues

### LINKED ISSUE #65: ECM fails broadcasting
When the transient state has a batch size of 1, but the A-SOH has a larger batch size, an error is thrown from the`SOCInterpolatedHealth` class. I haven't checked all cases, but it is possible that this happens the other way around too.

Steps to reproduce: 
```python
from moirae.models.ecm import EquivalentCircuitModel as ECM
from moirae.models.ecm import ECMASOH, ECMTransientVector, ECMInput

# Create a model
ecm = ECM()

# Create inputs
input0 = ECMInput(time=0, current=0)
input1 = ECMInput(time=0.1, current=2)

transient = ECMTransientVector.provide_template(has_C0=False, num_RC=0)
asoh = ECMASOH.provide_template(has_C0=False, num_RC=0)
asoh.mark_updatable(name='h0.base_values')
# Make a fake bacth
H0_vals = np.arange(5).reshape((5,1))
asoh.update_parameters(H0_vals)
print(f'A-SOH batch size: {asoh.batch_size}')  # A-SOH batch size: 5
print(f'Updatable parameters: {asoh.get_parameters()}')  # Updatable parameters: [[0] [1] [2] [3] [4]]

ecm.update_transient_state(previous_inputs=input0, new_inputs=input1, transient_state=transient, asoh=asoh)
```

Error message (last step in traceback, but the previous one is coming from the hysteris `get_value(soc=soc_k)` call in the ECM init line 88):
```python
File ~/some_path/auto-soh/moirae/models/ecm/utils.py:64, in SOCInterpolatedHealth.get_value(self, soc)
     [62]     if soc_batch_size > 0 and batch_size == 1:
     [63]         return np.repeat(y, soc.size, axis=0).reshape(input_dims)
---> [64]     return y.reshape(input_dims)
     [66] # Run the interpolator, but the results mean something different
     [67] y = self._interp_func(soc)  # interpolator adds a dimension

ValueError: cannot reshape array of size 5 into shape (1,1)
```

victorventuri:
This fixes #65. 

ECM can now properly broadcast regardless of whether transient vector and A-SOH are batched. That is, the model works when:

- Transient vector and A-SOH are both un-batched
- One is batched and the other is not (2 cases) 
- Both are batched (with equal batch_size)

Also provides more complete batch tests to account for these 4 cases.

### DIFF
--- a/moirae/models/ecm/__init__.py
+++ b/moirae/models/ecm/__init__.py
@@ -145,12 +145,13 @@ class EquivalentCircuitModel(CellModel):
         Vt = asoh.ocv(soc=transient_state.soc, temp=new_inputs.temperature)
 
         # Add I*R drop ('DCIR')
-        Vt += new_inputs.current * asoh.r0.get_value(soc=transient_state.soc,
-                                                     temp=new_inputs.temperature)
+        IR_drop = new_inputs.current * asoh.r0.get_value(soc=transient_state.soc,
+                                                         temp=new_inputs.temperature)
+        Vt = Vt + IR_drop
 
         # Check series capacitance
         if transient_state.q0 is not None:
-            Vt += transient_state.q0.copy() / asoh.c0.get_value(soc=transient_state.soc.copy())
+            Vt = Vt + transient_state.q0.copy() / asoh.c0.get_value(soc=transient_state.soc.copy())
 
         # Check RC elements
         if transient_state.i_rc.shape[-1] > 0:
@@ -160,9 +161,9 @@ class EquivalentCircuitModel(CellModel):
                  for rc in asoh.rc_elements]
             )  # Shape: (rc_rs, batch_dim, 1 resistance)
             V_drops = transient_state.i_rc * rc_rs[:, :, 0].T
-            Vt += np.sum(V_drops, axis=1, keepdims=True)
+            Vt = Vt + np.sum(V_drops, axis=1, keepdims=True)
 
         # Include hysteresis
-        Vt += transient_state.hyst
+        Vt = Vt + transient_state.hyst
 
         return ECMMeasurement(terminal_voltage=Vt)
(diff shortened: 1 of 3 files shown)


moirae/models/ecm/__init__.py
```diff
@@ -145,12 +145,13 @@ def calculate_terminal_voltage(
...
-        Vt += new_inputs.current * asoh.r0.get_value(soc=transient_state.soc,
-                                                     temp=new_inputs.temperature)
+        IR_drop = new_inputs.current * asoh.r0.get_value(soc=transient_state.soc,
+                                                         temp=new_inputs.temperature)
+        Vt = Vt + IR_drop
 
         # Check series capacitance
         if transient_state.q0 is not None:
-            Vt += transient_state.q0.copy() / asoh.c0.get_value(soc=transient_state.soc.copy())
+            Vt = Vt + transient_state.q0.copy() / asoh.c0.get_value(soc=transient_state.soc.copy())
```
WardLT, line 154: Minor performance optimization: I believe we can get rid of the `.copy()` on these, especially for `.q0`. Their values are not being altered so we need not protect them using copy.

moirae/models/ecm/utils.py
```diff
@@ -61,6 +61,8 @@ def get_value(self, soc: Union[Number, List, np.ndarray]) -> np.ndarray:
             y = self.base_values[:, 0]
             if soc_batch_size > 0 and batch_size == 1:
                 return np.repeat(y, soc.size, axis=0).reshape(input_dims)
+            elif soc_batch_size == 1 and batch_size > 1:
```
WardLT, line 64: Good catch!

WardLT (commented):
Looks good to me! If you think removing the `.copy` is sane, then let's try it before merging while our attention is on that part.

PR #71, opened 2024-08-20
Dual estimation framework done

victorventuri:
There are still some issues to iron out, but the logic appears correct.


moirae/estimators/online/dual.py
```diff
@@ -0,0 +1,172 @@
...
+
+from moirae.models.base import InputQuantities, OutputQuantities, GeneralContainer, HealthVariable, CellModel
+from .utils.model import CellModelWrapper, DegradationModelWrapper, convert_vals_model_to_filter, convert_numpy_to_model
+from moirae.estimators.online import OnlineEstimator
+from .filters.base import BaseFilter
+from .filters.distributions import MultivariateRandomDistribution, MultivariateGaussian
+from .filters.kalman.unscented import UnscentedKalmanFilter as UKF
+
+
+class individual_UKF_tuning_inputs(TypedDict):
```
WardLT, line 15: Style point: Classes should be named with PascalCase not snake_case. Could you change this and the next class to the capitalized style?

We should also move the "UKFTuningInputs" to the same module as the UKF to keep things grouped together better (e.g., if you want to use the class for the joint or other frameworks).

moirae/estimators/online/dual.py
```diff
@@ -0,0 +1,172 @@
...
+
+class dual_ukf_tuning_inputs(TypedDict):
+    transient: Optional[individual_UKF_tuning_inputs]
+    asoh: Optional[individual_UKF_tuning_inputs]
+
+
+class DualEstimator(OnlineEstimator):
+    """
+    In dual estimation, the transient vector and A-SOH are estimated by separate filters. This framework generally
+    avoids numerical errors related to the magnitude differences between values pertaining to transient quantities and
```
WardLT, line 29: Is the only purpose due to numerical issues? I figured the dual had something more to do with being able to reduce changes in the ASOH estimates by first attempting to explain measurement differences using transients.

moirae/estimators/online/dual.py
```diff
@@ -0,0 +1,172 @@
...
+from .filters.kalman.unscented import UnscentedKalmanFilter as UKF
+
+
+class individual_UKF_tuning_inputs(TypedDict):
+    alpha_param: Optional[float]
+    beta_param: Optional[float]
+    kappa_param: Optional[Union[float, Literal['automatic']]]
+
+
+class dual_ukf_tuning_inputs(TypedDict):
```
WardLT, line 21: Add docstring for these classes.

moirae/estimators/online/utils/model.py
```diff
@@ -23,17 +24,39 @@ def convert_vals_model_to_filter(
...
+
+    Args:
+        filter_array: array of numerical values to be converted
+        template: template object to be used; original object will not be modified
+        names: in case of a `HealthVariable`, can update specific names
+
+    Returns:
+        corresponding model-related object given by template with numerical values from the array
+    """
+    model_related_object = template.model_copy(deep=True)
```
WardLT, line 46: This method seems like it would make more sense as a method for the `GeneralContainer` and the `HealthVariable` class for two reasons:

1. The logic is very specific to the type of the input. There is only one line in common between the two types, and one of three arguments is not applicable to GeneralContainer.
2. The logic is not very specific to the use of CellModels in Filters. There are many cases where someone wants a new copy of a container but with updated values?

Could you add this method to those two classes? Maybe call it "make_copy(self, new_values)"

moirae/estimators/online/utils/model.py
```diff
@@ -57,18 +80,158 @@ def __init__(self,
...
+
+    @property
+    def num_hidden_dimensions(self) -> int:
+        return self.num_transients
+
+    def update_hidden_states(self,
+                             hidden_states: np.ndarray,
+                             previous_controls: np.ndarray,
+                             new_controls: np.ndarray) -> np.ndarray:
+        """
```
WardLT, line 118: Omit the docstring here. Autodocumentation tools will retrieve the definition from the base class, so we don't want two nearly-identical copies.

moirae/estimators/online/utils/model.py
```diff
@@ -57,18 +80,158 @@ def __init__(self,
...
+        self._num_output_dimensions = self.cell_model.calculate_terminal_voltage(new_inputs=self.inputs,
+                                                                                 transient_state=self.transients,
+                                                                                 asoh=self.asoh).to_numpy().shape[1]
 
     @property
     def num_output_dimensions(self) -> int:
         return self._num_output_dimensions
 
 
+class CellModelWrapper(BaseCellWrapper):
```
WardLT, line 95: Could you remove the duplicated text ("Base link between...") from the docstring in this class and the next? I've set up our documentation to show the relationship between classes, which makes it quick to see that CellModelWrapper is a type of BaseCellWrapper that is the "Base link between..."

tests/estimators/online/test_dual_ukf.py
```diff
@@ -0,0 +1,129 @@
...
+        covariance_transient=np.eye(2),
+        covariance_asoh=np.atleast_2d(1)
+    )
+    assert ukf_dual.state_names == ('soc', 'hyst', 'r0.base_values')
+    assert ukf_dual.output_names == ('terminal_voltage',)
+    assert ukf_dual.control_names == ('time', 'current')
+
+
+def test_simplest_rint():
+    # Create simplest possible Rint model, with no hysteresis, and only R0
```
WardLT, line 40: It was my intent to have "Simple_Rint" lack hysteresis. Did I accidentally have it in there still?

moirae/estimators/online/utils/model.py
```diff
@@ -57,18 +80,158 @@ def __init__(self,
...
+
+        # Get output
+        measurements = self.cell_model.calculate_terminal_voltage(new_inputs=inputs,
+                                                                  transient_state=transients,
+                                                                  asoh=self.asoh)
+
+        return measurements.to_numpy()
+
+
+class DegradationModelWrapper(BaseCellWrapper):
```
WardLT, line 152: Could you make tests for these classes as well?

moirae/estimators/online/filters/kalman/unscented.py
```diff
@@ -123,14 +121,26 @@ def __init__(self,
         self.cov_w = covariance_process_noise.copy()
         self.cov_v = covariance_sensor_noise.copy()
 
-        # Finally, we can set the weights for the mean and covariance updates
+    @cached_property
+    def gamma_param(self) -> float:
+        return self.alpha_param * np.sqrt(self._aug_len + self.kappa_param)
+
+    @cached_property
```
WardLT, line 128: Nice use of cached_property!

moirae/estimators/online/dual.py
```diff
@@ -0,0 +1,172 @@
...
+                           initial_controls=initial_controls,
+                           covariance_process_noise=transient_covariance_process_noise,
+                           covariance_sensor_noise=covariance_sensor_noise)
+        asoh_filter = UKF(model=asoh_wrapper,
+                          initial_hidden=asoh_hidden,
+                          initial_controls=initial_controls,
+                          covariance_process_noise=asoh_covariance_process_noise,
+                          covariance_sensor_noise=covariance_sensor_noise)
+
+        if filter_args is not None:
```
WardLT, line 164: Having to use `setattr` worries me a bit. It's verbose and, I'm guessing, resulted you in changing the UKF constructor to have it handle receiving the tuning parameters after-the-fact. Could you instead add this to the definitions of the filters?

```python
trans_filter = UKF(..., **filter_args.get('transient', {})
```
This will retrieve the values of the transient parameters, if available, and pass them as kwargs to the UKF initializer.

You'll also need to ensure `transient_args` is not None.

```python
if transient_args is None:
     transient_args = {}
```

moirae/estimators/online/dual.py
WardLT: We should consider having all our future filters implementations use a similar style of coming packed with their own TypedDict. Kind of nice on the user to clearly see what kind of kwargs they'll need to pass to other constructors.

WardLT (changes requested):
The core logic looks solid, but I've got some stylistic/docstring concerns and ideas for where we should move some of the logic. Nice work!

moirae/estimators/online/dual.py
victorventuri: I do not know for sure, but the tldr for me is "yes." I guess it depends on the implementation/filter being used. Conceptually, the dual estimation framework is always inferior to the joint, as it cannot adequately capture correlations between transient and A-SOH parameters: it takes the posterior from one, assumes it to be "truth", and does what it has to with it. In the case of UKF, it should help with poor numerical conditioning (to be seen), but nothing else really. With other filters, it may have other benefits.

WardLT: Fascinating! More to ponder :)

moirae/estimators/online/dual.py
WardLT: Another route could be to make delayed delivery of some arguments to a filter something always supported. I could see that being useful as we build larger assemblies of filters

WardLT: Mind you, I'd suggest we follow my first proposal now. Messing with the filter seems like it's unnecessary now and, therefore, out of scope for this PR

moirae/estimators/online/dual.py
victorventuri: Good idea! Moved this class to the filter location and renamed it.

moirae/estimators/online/dual.py
victorventuri: One other useful example is: imagine we find out that UKF is great for updating transient states, but a particle filter works better for the A-SOH. The dual allows us to use them both separately, while the joint requires the same filter for everything.;

moirae/estimators/online/dual.py
victorventuri: Gotcha!

moirae/estimators/online/dual.py
victorventuri: Came up with a way to do the first suggestion, but would like you to look it over just to make sure @WardLT.

moirae/estimators/online/utils/model.py
victorventuri: Done!

moirae/estimators/online/utils/model.py
victorventuri: Done!

moirae/estimators/online/utils/model.py
victorventuri: Yep, taking care of it!

tests/estimators/online/test_dual_ukf.py
victorventuri: No, you don't. I am not using the simple_rint fixture here because I initialize some of the values to something other than the default ones. I figured that was easier than pulling simple_rint, making copies (for real and guess objects) and changing the values. Would you prefer if I used the simple_rint?

moirae/estimators/online/utils/model.py
victorventuri: On it!

tests/estimators/online/test_dual_ukf.py
WardLT: Nah. Just wanted to make sure I didn't make the "simple int" too complicated. If it doesn't work here, I don't oppose using something else

moirae/estimators/online/dual.py
WardLT: Nice! I hadn't seen NotRequired before. It makes a lot of sense to do it that way

WardLT (approved):
Looks good to me. Feel free to "squash and merge" when you're ready to move forward!

Issue #59, opened 2024-08-14
Joint UKF notebook shows numerical issues with Thenevin example


victorventuri:
When running the second example in the joint UKF demo notebook, numerical issues show up: somewhere along the way, a covariance matrix became not positive semi-definite, likely as a result of the large difference in order of magnitude between values in transient vectors and A-SOH parameters. 

We should investigate if re-adding the A-SOH normalization could help with this. 

This is something that, while less relevant, could be useful for the dual estimation as well, so maybe there is a way in which we can implement only once, and both frameworks have access to it. Tagging @WardLT to discuss further.

WardLT:
Do you think we can make a "NormalizerFilter" wrapper over a BaseFilter? 

I'm picturing a metaclass of BaseFilter which a filter it wraps and a list of normalization factors, then handles adjusting the hidden states?

victorventuri:
This is a great idea! I had not imagined the normalization sitting on the `filters` side of it, but that seems like it would be an easier way of making sure both joint and dual frameworks interact with it in the same way :thinking:

victorventuri:
@WardLT this is also happening in the dual, which is unexpected... I will do some more checks to see if the problem is not elsewhere, but perhaps we should consider prioritizing the normalization... will keep you posted

WardLT:
I was thinking about the normalization, and have a proposal: a wrapper over a filter which transforms data being passed to the underlying filter.

 The problem I missed before was that the filter makes calls to the physics model, and that we must denormalize before the points before calling the physics model. So, there must also be something to handle normalization for the model. The normalizing filter wrapper will also have a to apply a such a wrapper over the model of the underlying filter.

PS. This needs a picture. I'll sketch one tomorrow. Just wanted to drop this note while it was on the brain.

victorventuri:
As I am giving this some more thought, i will jot down ideas here. At first, there are two questions to consider:

1. Who needs to know that something (the A-SOH) is normalized?
2. Where does the normalization implementation live? 

I feel the answer to the first question guides the answer to the second question. Based on that, I discuss three candidate classes/objects that may need knowledge of normalization:

- `HealthVariable`: while i am open to discussion if you find this is viable, I am *a priori* strongly opposed to this because
    - Pros:
        - little to nothing on the estimator side would need to be changed
    - Cons: 
        - the model deals with "raw" values, and normalization is meaningless there
[...]

PR #72, opened 2024-08-22
Proposal to allow ModelWrapper and Filter to operate in different coordinate systems

WardLT:
A proposal to help with #59

- [ ] Split off transform to its own class
- [ ] Introduce transform class as argument to ModelWrapper
- [ ] Update the MvRD to have a transform operation which produces a new distribution after applying a transform
- [ ] Allow the `get_state` function of the estimator to detect when the state used by an underlying filter requires a transformation


victorventuri:
Jotting down some notes here: 

- the `TransformedModel` class has methods like `transform_covariance_to_wrapped`, which implies that it knows at least what the underlying probability distribution is. 
    - at first, this doesn't seem like quite a problem, as all pdfs have a covariance, but the problem is that it becomes the responsibility of the `TransformedModel` to know how to update that
    - to me, this clashes with the modularity idea: the wrappers should similarly be agnostic to the `BaseFilter` being used, and be able to work with standard KF, extended KF (neither of these implemented), UKF, particle filters, RNN-based filters, etc. 
- given we are creating this functionality in the `TransformedModel`, I feel like the `TransformedFilter` is superfluous: its only "new" functionality is to return the state de-transformed, but that can be done at the `OnlineEstimator` level, since that has access to the necessary functions, and is the user-facing object.
- I see you have a `TODO (wardlt): Move this whole module to the ./filters/ package?`, which is a sentiment i agree with (though maybe this was a leftover from an older branch?). since the `ModelWrapper` is the way the `BaseFilter`s expect to interact with models, I liked having it defined in the `filters` directory

WardLT:
> The TransformedModel class has methods like transform_covariance_to_wrapped, which implies that it knows at least what the underlying probability distribution is.
Yea, that is a bit of a problem. I'm eager to see your thoughts on how to make the MVP classes allow transformations. 

> ... given we are creating this functionality in the TransformedModel, I feel like the TransformedFilter is superfluous: its only "new" functionality is to return the state de-transformed, but that can be done at the OnlineEstimator level, since that has access to the necessary functions, and is the user-facing object.
Agreed. The primary user of the `TransformedFilter` is so that the OnlineEsimator code need not be changed to allow for transformation, but perhaps there's a simple way to do that. 

> I see you have a TODO (wardlt): Move this whole module to the ./filters/ package?, which is a sentiment i agree with (though maybe this was a leftover from an older branch?)
It's actually something new that came up while I was trying to consolidate the model-wrapper-related stuff to the same module, and also to promote it from a generic "utils" given it's large importance in implementing a filter.

victorventuri:
Ok, so I came up with something for the `MvRD`. Looking forward to sharing it with you in a bit, and getting your feedback. I was doing it in the `demonstrate_ignorance.ipynb` notebook (which is naturally ignored) to test it out, since i didn't feel like making a full new branch for it, given there's a good chance it would go nowhere. I think once we sync up today, we should just create a `asoh_normalization` branch to work on whichever solution we come up with. What do you think?

WardLT:
Let's do it :)

WardLT:
Closed in favor of a different branch

PR #73, opened 2024-08-26
Estimator normalization

victorventuri:
`ModelWrapper`s now accept user-specified transforms to help convert between filter and model coordinate systems. This has been applied and tested to the joint estimation framework, and similar changes on the dual framework will come later.


moirae/estimators/online/filters/base.py
```diff
@@ -1,21 +1,104 @@
...
     """
+    def __init__(self,
+                 hidden_conversion_operator: Optional[ConversionOperator] = IdentityConversionOperator(),
+                 control_conversion_operator: Optional[ConversionOperator] = IdentityConversionOperator(),
+                 output_conversion_operator: Optional[ConversionOperator] = IdentityConversionOperator()) -> None:
+        self._hidden_conversion = hidden_conversion_operator.model_copy(deep=True)
+        self._control_conversion = control_conversion_operator.model_copy(deep=True)
+        self._output_conversion = output_conversion_operator.model_copy(deep=True)
+
+    def _convert_from_hidden_samples(self, filter_hidden_samples: np.ndarray) -> np.ndarray:
```
WardLT, line 49: Can we do without these functions? They are just wrappers over another component (_hidden_conversion) that implementations have access to anyway.

moirae/estimators/online/filters/conversions.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+class IdentityConversionOperator(ConversionOperator):
+    """
+    Class that implements simple identity operation, that is, it does not change the inputs.
+    """
+    def transform_samples(self, samples: Optional[np.ndarray]) -> np.ndarray:
+        if samples is None:
+            return
+        return samples.copy()
```
WardLT, line 87: Why are we making copies? There should not be a worry about modifying the original data.

moirae/estimators/online/filters/conversions.py
```diff
@@ -0,0 +1,191 @@
...
+        """
+        raise NotImplementedError('Implement in child class!')
+
+
+class IdentityConversionOperator(ConversionOperator):
+    """
+    Class that implements simple identity operation, that is, it does not change the inputs.
+    """
+    def transform_samples(self, samples: Optional[np.ndarray]) -> np.ndarray:
+        if samples is None:
```
WardLT, line 85: The base class's definition says samples is not optional, so let's not change it in the subclass. Also, why would the only argument to a function be optional?

moirae/estimators/online/filters/conversions.py
```diff
@@ -0,0 +1,191 @@
...
+            return
+        return samples.copy()
+
+    def transform_covariance(self, covariance: Optional[np.ndarray]) -> np.ndarray:
+        if covariance is None:
+            return
+        return covariance.copy()
```
WardLT, line 92: As above, let's make the functions none-optional.

moirae/estimators/online/filters/conversions.py
```diff
@@ -0,0 +1,191 @@
...
+            return
+        return transformed_samples.copy()
+
+    def inverse_transform_covariance(self, transformed_covariance: Optional[np.ndarray]) -> np.ndarray:
+        if transformed_covariance is None:
+            return
+        return transformed_covariance.copy()
+
+
+class LinearConversionOperator(ConversionOperator, validate_assignment=True):
```
WardLT, line 105: Let's avoid "validate_assigment" wherever possible. It adds a bunch of overhead

moirae/estimators/online/filters/conversions.py
```diff
@@ -0,0 +1,191 @@
...
+    """
+    Class that implements a linear function as a transformation (strictly speaking, this is not a linear transformation,
```
WardLT, line 107: Could you remove the parenthetical? I think it's too verbose and confusing

moirae/estimators/online/filters/conversions.py
```diff
@@ -0,0 +1,191 @@
...
+        Ensures the multiplicative array is stored as a 2D array (or as a shapeless object)
+        """
+        shape = multi.shape
+        if len(shape) > 2:
+            raise ValueError(f'Additive factor must be at most 2D, but, instead, is has shape {shape}!')
+        if len(shape) == 1:
+            return np.diag(multi)
+        return multi
+
+    @computed_field
```
WardLT, line 146: Could you change this from "computed_field," which adds the value to the serialized version of the class to "cached_property?" We don't need the additional functionality which comes with storing the values in the JSON form of this class.

moirae/estimators/online/filters/distributions.py
```diff
@@ -43,12 +45,48 @@ def combine_with(self, random_dists: Iterable[Self]) -> Self:
...
+    @field_validator('mean', mode='after')
+    @classmethod
+    def mean_1d(cls, mu: np.ndarray) -> np.ndarray:
+        """ Making sure the mean is a vector """
+        mean_shape = mu.shape
+        if not mean_shape:
+            raise ValueError('Mean must be Sized and have a non-empty shape!')
+        if len(mean_shape) > 2 or (len(mean_shape) == 2 and 1 not in mean_shape):
+            raise ValueError('Mean must be a 1D vector, but array provided has shape ' + str(mean_shape) + '!')
+        elif len(mean_shape) == 2:
```
WardLT, line 84: Why not throw an error when a user provides an incompatible value? Adding more "implicit" operations like this type conversion below makes the code more complicated without specific benefits.

moirae/estimators/online/joint.py
```diff
@@ -56,24 +59,29 @@ def step(self, inputs: InputQuantities, measurements: OutputQuantities) -> \
         joint_estimate, output_predicted = self.filter.step(new_controls=refactored_inputs,
                                                             measurements=refactored_measurements)
 
-        return joint_estimate, output_predicted
+        return (joint_estimate.convert(conversion_operator=self.joint_model._hidden_conversion),
+                output_predicted.convert(conversion_operator=self.joint_model._output_conversion))
```
WardLT, line 63: Don't call private functions from another class. Make the interfaces that are required public.

moirae/estimators/online/joint.py
```diff
@@ -116,16 +137,24 @@ def initialize_unscented_kalman_filter(cls,
...
             joint_hidden_covariance = covariance_joint
+        # Convert covariance!
+        joint_hidden_covariance = asoh_normalizer.inverse_transform_covariance(
+            transformed_covariance=joint_hidden_covariance)
         joint_initial_hidden = MultivariateGaussian(mean=joint_hidden_mean[0, :], covariance=joint_hidden_covariance)
         # Initial controls
-        initial_controls = convert_vals_model_to_filter(model_quantities=initial_inputs,
-                                                        uncertainty_matrix=inputs_uncertainty)
+        initial_controls = convert_vals_model_to_filter(
+            model_quantities=joint_model._convert_to_control_samples(model_control_samples=initial_inputs),
```
WardLT, line 146: As above, do not call private methods and let's also not have functions that are just 1:1 mappings to calling a function from another class.

tests/estimators/online/filters/test_conversions.py
```diff
@@ -0,0 +1,188 @@
+import numpy as np
+
+from moirae.estimators.online.filters.conversions import IdentityConversionOperator, LinearConversionOperator
+
+
+def test_identity_operator():
```
WardLT, line 6: Nice test!

tests/estimators/online/test_joint_ukf.py
```diff
@@ -266,3 +266,176 @@ def test_joint_ecm() -> None:
         assert vt_sigs[i - 1] >= gaussian_metrics[i] / 100, \
             'Real Vt value is within %d std of prediction only %2.2f%% of the time, rather than %2.0f%%!' % \
             (i, 100 * qt_sigs[i - 1], gaussian_metrics[i])
+
+
+def test_joint_ecm_with_normalization() -> None:
```
WardLT, line 271: Not something to change now, but advice for later: let's avoid copying-and-pasting code between these test functions. The code here is mostly identical for the test without normalization, so let's separate the logic that's different (e.g., making the estimator) and re-use the bulkly testing logic at the end.

moirae/estimators/online/filters/base.py
```diff
@@ -1,21 +1,104 @@
...
+    Args:
+        hidden_conversion_operator: operator that determines how the filter hidden states must be converted to be given
+            to the model
+        control_conversion_operator: operator that determines how filter controls must be converted to be given to the
+            model
+        output_conversion_operator: operator that determines how model outputs must be converted to be given to the
+            filter
     """
+    def __init__(self,
+                 hidden_conversion_operator: Optional[ConversionOperator] = IdentityConversionOperator(),
```
WardLT, line 42: These are not actually optional, given that all of the subsequent code assumes them to be not None.

moirae/estimators/online/filters/base.py
```diff
@@ -1,21 +1,104 @@
...
+                 control_conversion_operator: Optional[ConversionOperator] = IdentityConversionOperator(),
+                 output_conversion_operator: Optional[ConversionOperator] = IdentityConversionOperator()) -> None:
+        self._hidden_conversion = hidden_conversion_operator.model_copy(deep=True)
```
WardLT, line 45: These conversion operations must be available to calling classes (e.g., the Estimator), so mark them as public by removing the underscore in front and documenting them as class attributes.

WardLT (changes requested):
The two things that must be changed:
1. Remove the functions which just call the underlying conversion operators in `ModelWrapper` and instead make the conversion tools public. Why create two interfaces (the original converter, the wrapped calls to the converted) which do the same thing?
2. Make the validators more opinionated. Throw an error rather than trying to assume user intent if they pass an incorrectly-shaped value.

moirae/estimators/online/filters/base.py
victorventuri: Yep, killing them

moirae/estimators/online/filters/conversions.py
victorventuri: I thought I had a reason, but I cannot remember it now, so it's probably not important. Based on good code design, if the reason is important, we will have to find a different work around later, because, as you said, this should not be optional. Fixing it.

moirae/estimators/online/filters/conversions.py
victorventuri: True. Fixed.

moirae/estimators/online/filters/conversions.py
victorventuri: Sounds good!

moirae/estimators/online/filters/conversions.py
victorventuri: Sure thing!

moirae/estimators/online/filters/conversions.py
victorventuri: Sure, I had added it here because it caches it, but there are other ways of doing the same thing without the pydantic overhead. Taking care of it.

moirae/estimators/online/filters/distributions.py
victorventuri: I added here because, in the model-related objects (`HealthVariable` and `GeneralContainer`), things are specified as 2D arrays, even when they are really 1D, and I didn't want to end up having a situation where everything breaks because somewhere we forgot to add a flatten. Based on this, do you still think I should throw an error here?

moirae/estimators/online/joint.py
victorventuri: Of course, idk what I was thinking. Fixing these now.

tests/estimators/online/test_joint_ukf.py
victorventuri: That's very good. I was lazy here; should have made these a fixture. I just wanted to make sure things were working, but you are 100% right. Let's take care of fixing this later, and I will make sure to not mess up with the Dual

moirae/estimators/online/joint.py
victorventuri: Fixed.

moirae/estimators/online/filters/distributions.py
WardLT: Yes, so long as we make it clear what they did wrong. Maybe even add "did you forget "flatten"?"

victorventuri: Gotcha, taking care of it!

moirae/estimators/online/utils/model.py
```diff
@@ -45,7 +45,11 @@ def __init__(self,
                  cell_model: CellModel,
                  asoh: HealthVariable,
                  transients: GeneralContainer,
-                 inputs: InputQuantities):
+                 inputs: InputQuantities,
+                 converters: Optional[ModelWrapperConverters] = ModelWrapperConverters.defaults()):
```
WardLT, line 49: Another "Optional" that cannot be None. I'll fix it before merging

WardLT (approved):
Looks great! I'm going to change one thing (a stray Optional) then merge if the tests pass

moirae/estimators/online/utils/model.py
victorventuri: oh sorry, I guess I missed this one... thanks!

PR #75, opened 2024-08-27
Dual normalization

victorventuri:
Implements normalization in dual framework, and provides an easy way to initialize an A-SOH-normalized dual UKF. Related to discussions in #59.

### DIFF
--- a/moirae/estimators/online/dual.py
+++ b/moirae/estimators/online/dual.py
@@ -11,6 +11,7 @@ from .filters.base import BaseFilter
 from .filters.distributions import MultivariateRandomDistribution, MultivariateGaussian
 from .filters.kalman.unscented import UnscentedKalmanFilter as UKF
 from .filters.kalman.unscented import UKFTuningParameters
+from .filters.conversions import LinearConversionOperator, IdentityConversionOperator
 
 
 class DualUKFTuningParameters(TypedDict):
@@ -99,7 +100,7 @@ class DualEstimator(OnlineEstimator):
 
         # Translate to model object
         estimated_transient = self.transients.make_copy(values=transient_converted.get_mean())
-        estimated_asoh = self.asoh_wrapper._convert_hidden_to_asoh(hidden_states=asoh_converted.get_mean())
+        estimated_asoh = self.asoh.make_copy(values=asoh_converted.get_mean())
         return estimated_transient, estimated_asoh
 
     def step(self, inputs: InputQuantities, measurements: OutputQuantities) -> \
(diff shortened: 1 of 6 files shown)


moirae/estimators/online/dual.py
```diff
@@ -61,20 +66,41 @@ def __init__(self, transient_filter: BaseFilter, asoh_filter: BaseFilter) -> Non
...
+                :class:`~moirae.estimators.online.filters.distributions.MultivariateRandomDistribution` in
+                model-coordinates
+        """
         # We need to get the hidden states on both filters
         transient_hidden = self.trans_filter.hidden.model_copy(deep=True)
         asoh_hidden = self.asoh_filter.hidden.model_copy(deep=True)
-        return transient_hidden.combine_with(asoh_hidden)
+
+        # We need to transform them accordingly
+        transient_converted = transient_hidden.convert(conversion_operator=self.trans_filter.model.hidden_conversion)
```
WardLT, line 86: Do we need the copy anymore if the the `convert` method produces a new instance of the distribution?

moirae/estimators/online/dual.py
```diff
@@ -149,27 +197,51 @@ def initialize_unscented_kalman_filter(
...
-                           covariance_process_noise=transient_covariance_process_noise,
-                           covariance_sensor_noise=covariance_sensor_noise,
-                           **filter_args.get('transient', UKFTuningParameters.defaults()))
-        asoh_filter = UKF(model=asoh_wrapper,
-                          initial_hidden=asoh_hidden,
-                          initial_controls=initial_controls,
-                          covariance_process_noise=asoh_covariance_process_noise,
-                          covariance_sensor_noise=covariance_sensor_noise,
-                          **filter_args.get('asoh', UKFTuningParameters.defaults()))
+        trans_filter = UKF(
```
WardLT, line 228: Tests which ensure we didn't have these numerous conversions backwards would be awesome

WardLT (commented):
A few comments, since it sounds like you're still working on this

moirae/estimators/online/dual.py
victorventuri: You are right, no need to crowd memory. What do you think of this new version? (btw, I checked to make sure they run fine in the slow tests locally)

moirae/estimators/online/dual.py
victorventuri: Good idea! Will write one as well!

moirae/estimators/online/dual.py
WardLT: I was more pointing out the `model_copy` statements in the previous lines. You shouldn't need to copy and then convert, just convert.

WardLT (approved):
Looks great!

PR #85, opened 2024-09-11
Degradation model

victorventuri:
Added degradation model to models, simulator, and estimators. 

Had to fix minor bugs in the ECM model:

1. In the `SOCInterpolatedHealth`, in cases were the underlying `HealthVariable` was not batched, the interpolation accidentally changed the values in `base_values` attribute. This problem only showed up when dealing with degradation because we use the same A-SOH object after updating the transient vector in the `BaseCellModelWrapper`s 
2. Still in the `SOCInterpolatedHealth`, a problem occurred when the `base_values` were batched and the SOC was passed as an unbatched 2D array of shape `(1,1)`: the output was a 3D array of shape `(batch_size, 1, 1)`, which is not the desired behavior. This PR also corrects that to make sure the output of the `get_value()` method is always a 2D array. 

Re-ran tests (including slow ones) and all passed. Re-ran relevant demo notebooks as well. Things seem to be working with a dummy degradation model (which does not degrade the A-SOH)


moirae/estimators/online/utils/model.py
```diff
@@ -32,6 +33,14 @@ class BaseCellWrapper(ModelWrapper):
     """
     Base link between the :class:`~moirae.model.base.CellModel` and the numpy-only interface of
     the filter implementations.
+
+    Args:
+        cell_model: Model which defines the physics of the system being modeled
```
WardLT, line 38: Should we include the degradation model wrapper as a part of every CellModel wrapper? I could imagine wanting to see the effect of degradation even if I'm not adjusting its parameters.

moirae/models/utils.py
```diff
@@ -0,0 +1,12 @@
...
+Collection of utility functions and classes for the models
+"""
+from .base import DegradationModel, HealthVariable
+
+
+class DummyDegradation(DegradationModel):
+    """
+    Class corresponding to no degradation, that is, the updated A-SOH is the same as the previous A-SOH.
+    """
+    def update_asoh(self, previous_asoh: HealthVariable, *args, **kwargs) -> HealthVariable:
```
WardLT, line 11: Could you copy the function signature from the base class? It should not change for subclasses.

moirae/models/utils.py
```diff
@@ -0,0 +1,12 @@
...
+        return previous_asoh.model_copy(deep=True)
```
WardLT, line 12: Why are we copying here? That feels like an unnecessary computational overhead, especially if we do it every step.

moirae/models/utils.py
```diff
@@ -0,0 +1,12 @@
+"""
+Collection of utility functions and classes for the models
+"""
+from .base import DegradationModel, HealthVariable
+
+
+class DummyDegradation(DegradationModel):
```
WardLT, line 7: Dummy feels nonspecific. What about "NoDegradation"?

tests/models/test_utils.py
```diff
@@ -0,0 +1,40 @@
...
+    Creates a very simple health variable
+    """
+    d0 = SubHeathVariable(x=5)
+    d1 = SubHeathVariable(x=-5)
+    ex = ExampleHealthVariable(d=(d0, d1))
+    ex.mark_all_updatable()
+    return ex
+
+
+def test_dummy_deg(example_hv) -> None:
```
WardLT, line 37: Why is this test in a new file? That seems unnecessary, especially if it results in copy-pasting data.

I'd suggest either combining the tests into a single file or making the "example_hv" fixture part of a "conftest.py" file so that it can be used in multiple test files.

moirae/models/ecm/utils.py
```diff
@@ -58,7 +58,7 @@ def get_value(self, soc: Union[Number, List, np.ndarray]) -> np.ndarray:
 
         # Special case: no interpolation
         if self.base_values.shape[-1] == 1:
-            y = self.base_values[:, 0]
+            y = self.base_values[:, 0].copy()
```
WardLT, line 61: Why the copy here? It doesn't look any any of the subsequent actions alter values of y.

moirae/models/ecm/utils.py
```diff
@@ -73,6 +73,8 @@ def get_value(self, soc: Union[Number, List, np.ndarray]) -> np.ndarray:
         if soc_batch_size > 1 and batch_size > 1:
             y = np.diag(y.squeeze())  # Match the SOC with the model its calling
         elif soc_batch_size == 1 and batch_size > 1:
+            if len(input_dims) > 1:
```
WardLT, line 76: Do you have a simple test which breaks before making this change? It'd be great to add one so that we ensure any future modifications (not that I have them in mind, it's just good to be prepared) don't revert to bad behavior.

WardLT (changes requested):
Almost ready for merging. The big issues are:
1. Avoid copying. Both a few unnecessary `.copy` statements and code copy-pasted between files.
2. If you agree, make degradation models part of every CellModelWrapper and the base class. It seems like we should always be running with degredation.

moirae/models/ecm/utils.py
victorventuri: The subsequent actions don't, but what we do with the values after does. The problem is that `np.reshape` does not return a copy of the array, but a different view of it. For example:
```python
a = np.arange(10)
print(a)  # shows [0 1 2 3 4 5 6 7 8 9]
b = a.reshape((5,2))
b *= 5
print(a)  # shows [ 0  5 10 15 20 25 30 35 40 45]
```
So, what that means is that what we are actually returning here is something that has direct access to the base values, which is not what we want. 

In the `ECM().update_transient_states` method, we get the value of the hysteresis and store it in variable `M`, but then we multiply that by the current, and that changes the `base_values` of the `asoh.h0` attribute.

moirae/estimators/online/utils/model.py
victorventuri: The `CellModelWrapper` is not allowed to touch the A-SOH, so I don't think the `DegrdationModel` should be included there. I don't think I understood the reason why we should include it here...

moirae/models/ecm/utils.py
victorventuri: Yes, so the problem is that, in the `calculate_terminal_voltage` method inside the ECM, we are generally expecting arrays to be returned as 2D, with the first dimension being `batch_size`, and the other one, some parameterization-related dimension (in most cases, it should be 1). 

However, in this function, when the SOC is provided as a `(1,1)`-shaped array, and interpolation is needed, the returned vector is 3D, with shape `(batch_size,1,1)`. So, when we go forward to a next step that, for example, doesn't need interpolation, and returns an array of shape `(batch_size,1)`, numpy's broadcasting sums these together into an array of shape `('batch_size, batch_size,1)`, which cannot be properly squeezed. 

Here is an example of this: create a Thenevin model in which `R0` needs interpolation, but the resistor in the `RC` does not, but both are batched:
```python
# Define the values of R0 to use
r0_vals = np.array([5,4,6])

# Get A-SOH and set R0 values and interpolation style
asoh = ECMASOH.provide_template(has_C0=False, num_RC=1, R0=r0_vals)
asoh.r0.interpolation_style = 'quadratic'

# Mark R0 and RC as updatable
asoh.mark_updatable(name='r0.base_values')
asoh.mark_updatable(name='rc_elements.0.r.base_values')
# asoh_rint.mark_updatable(name='rc_elements.0.c.base_values')

# Now, create a batch of size 4
r0_batch = np.arange(12).reshape((4,3))
r_rc_batch = (1 + np.arange(4).reshape((4,-1)))
asoh.update_parameters(values=np.hstack((r0_batch, r_rc_batch)))

# Set up transient and input
transient = ECMTransientVector.provide_template(has_C0=False, num_RC=1, soc=1.0)
control = ECMInput(time=0., current=1.)

# Get terminal voltage
out = ECM().calculate_terminal_voltage(new_inputs=control,
                                       transient_state=transient,
                                       asoh=asoh)
```

The last line will raise a `pydantic` `ValidationError` from the `ScalarParameter` because the dimensions of the array are `(4, 4, 1)`:
```python
ValidationError: 1 validation error for ECMMeasurement
terminal_voltage
  Value error, Inconsistent dimensionality. Trying to store a 2 array as a scalar [type=value_error, input_value=array([[[ 5.7944439],
   ...
        [14.7944439]]]), input_type=ndarray]
```

moirae/models/utils.py
victorventuri: Sounds good, fixed!

moirae/models/utils.py
victorventuri: Sorry, reverted to bad habits here. Fixing it!

moirae/models/utils.py
victorventuri: Ultimately, because we do not know what is going to be done with the A-SOH object returned. My original idea is that the `DegradationModel` always outputs a new `HealthVariable` object that represents a new state of the A-SOH, and it's up to whoever is using it to decide what to do with it and with the original A-SOH. I feel like this is a "better safe than sorry" type of overhead (kind of similar to the `SOCInterpolatedHealth` one), but I am open to changing it. I just feel like this particular overhead is relatively minor compared to the other things we are doing.

WardLT: That's fair. Returning a new copy makes sense to me

moirae/models/ecm/utils.py
WardLT: Nice catch!

moirae/models/ecm/utils.py
WardLT: Can you reproduce that error with a simple test which uses only the SOC interpolation? I think I see the problem, but let's make sure we can express it in a simple form and add a test so that we can ensure it never becomes un-fixed.

WardLT (approved):
I'm fine with the modifications. Please add a test for the new SOC dimensionality issue then feel free to merge

moirae/models/ecm/utils.py
victorventuri: Thank you! Resolving it.

moirae/models/ecm/utils.py
victorventuri: Yep, adding a test on the `test_components.py`

PR #86, opened 2024-09-12
Improve documentation for cell models

WardLT:
Cell models and all of their related components.

### DIFF
--- a/docs/extending.rst
+++ b/docs/extending.rst
@@ -1,0 +1,26 @@
+Extending Moirae
+================
+
+Welcome to our in progress page on how to add new features to Moirae.
+It will become more organized as the package becomes more mature.
+For now, it is disconnected sections written as we build capabilities.
+
+Creating a Health Variable
+--------------------------
+
+Define a new system state by subclassing ``HealthVariable`` then providing
+adding attributes which describe the learnable parameters of a system.
+
+Attributes which represents a health parameter must be numpy arrays,
+other ``HealthVariable`` classes,
+or lists or dictionaries of other ``HealthVariable`` classes.
+
+The numpy arrays used to store parameters are 2D arrays where the first dimension is a batch dimension,
+even for parameters which represent scalar values.
+Use the :class:`ScalarParameter` type for scalar values and :class:`ListParameter` for list values
+to enable automatic conversion from user-supplied to the internal format used by :class:`HealthVariable`.
+
+.. note::
+
+    The ``ListParameter`` and ``ScalarParameter`` classes also supply methods needed for serialization to
+    and parsing form JSON.
--- a/docs/extending.rst
+++ b/docs/extending.rst
@@ -8,9 +8,8 @@ For now, it is disconnected sections written as we build capabilities.
 Creating a Health Variable
 --------------------------
 
-Define a new system state by subclassing ``HealthVariable`` then providing
-adding attributes which describe the learnable parameters of a system.
-
+Define the parameters for a new battery system by
+subclassing :class:`~moirae.models.base.HealthVariable`.
 Attributes which represents a health parameter must be numpy arrays,
 other ``HealthVariable`` classes,
 or lists or dictionaries of other ``HealthVariable`` classes.
--- a/docs/system-models.rst
+++ b/docs/system-models.rst
@@ -148,9 +148,29 @@ or omit the specific names to set all updatable variables
     assert model.updatable_names == ['resistance.full', 'resistance.empty']
     model.update_parameters([[0.2, 0.1]])  # As a (1, 2) array for 1-sized batch of 2 values
 
-Cell Model
-----------
+Defining the Cell Physics
+-------------------------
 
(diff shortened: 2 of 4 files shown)


docs/system-models.rst
```diff
@@ -53,16 +54,123 @@ necessary to specify a particular storage systems.
...
+
+Setting Values of Parameters
+++++++++++++++++++++++++++++
+
+Provide a list of new values and a list of names to the ``update_parameters`` function.
+
+.. code-block:: python
+
+    model.updatable.add('resistance')  # Allows resistance fields to be updated
+    model.update_parameters([[0.1]], ['resistance.full'])
```
victorventuri, line 142: Can we include the keyword arguments? I want to encourage users to always use them, instead of relying solely on positional args.

docs/system-models.rst
```diff
@@ -53,16 +54,123 @@ necessary to specify a particular storage systems.
...
+    assert model.updatable_names == ['resistance.full', 'resistance.empty']
+    model.update_parameters([[0.2, 0.1]])  # As a (1, 2) array for 1-sized batch of 2 values
+
+Defining the Cell Physics
+-------------------------
+
+All storage systems are represented using a :class:`~moirae.models.base.CellModel`
+that provides two functions:
+
+1. updating transient state given inputs, and
```
victorventuri, line 157: It also needs the A-SOH and previous inputs, so should we just say "1. updating transient states and, 2. calculating terminal voltage" for conciseness?

docs/system-models.rst
```diff
@@ -53,16 +54,123 @@ necessary to specify a particular storage systems.
...
+
+.. ::
+
+    We still need to....
 
-All storage systems are represented using a ``CellModel``.
-The model class defines how the transient states change given
-an inputs applied to a system defined using certain health parameters.
+    1. Describe where any parameters for the degredation model come from
+    2. Indicate if there are any additional states held by the degradation model
```
victorventuri, line 175: The short answer is yes. But, which states, will be dependent on the degradation model we are implementing...

docs/system-models.rst
```diff
@@ -53,16 +54,123 @@ necessary to specify a particular storage systems.
...
 and the ability annotate which parameters are treated as updatable.
 
-.. note:: TODO, figure out how much documentation to copy over from the class docstring
+Consider the following parameter set as an example
 
-Models
-------
+.. code-block:: python
+
+    class Resistance(HealthVariable):
```
victorventuri, line 66: Should we make it clear that this is not how the implemented `Resistance` works?

victorventuri (approved):
Nice work!

docs/extending.rst
```diff
@@ -0,0 +1,26 @@
...
+
+Welcome to our in progress page on how to add new features to Moirae.
+It will become more organized as the package becomes more mature.
+For now, it is disconnected sections written as we build capabilities.
+
+Creating a Health Variable
+--------------------------
+
+Define a new system state by subclassing ``HealthVariable`` then providing
+adding attributes which describe the learnable parameters of a system.
```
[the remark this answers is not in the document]
WardLT: Good points. I'll clean up this page

docs/system-models.rst
```diff
@@ -6,12 +6,13 @@ A battery system, regardless of its scale or technology, is represented by...
 - A **transient state** representing how the parameters which vary predictably given operating conditions, like the state of charge
 - An **advanced state of health (ASOH)** representing the parameters which vary slowly and unpredictably, like the internal resistance
 - A set of **input quantities** and **output quantities** that define how a system is used and what signals it produces.
-- A **model** describing the transient state evolves over time given the ASOH and inputs.
+- A **cell model** describing the transient state evolves over time given the ASOH and inputs.
```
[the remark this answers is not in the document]
WardLT: Yup 😮‍💨

docs/system-models.rst
```diff
@@ -53,16 +54,123 @@ necessary to specify a particular storage systems.
...
+
+or indirectly using :meth:`get_parameters`.
+
+.. code-block:: python
+
+    assert np.allclose(model.get_parameters(['resistance.full']), [[0.2]])
+
+The name of a variable within a hierarchical health variable contains the path to its submodel
+and the name of the attribute of the submodel separated by periods.
+For example, the resistance at fully charge is "resistance.full".
```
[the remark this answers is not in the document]
WardLT: Thanks. I missed that in proofreading

docs/system-models.rst
WardLT: Sounds great. How about we keep terminal voltage as an example because there could be other observables eventually

docs/system-models.rst
WardLT: Agreed. FYI, these don't comments are just intended for us later and don't render in the HTML version

docs/system-models.rst
WardLT: I don't think that's necessary given how different it is from the resistance we have now

PR #91, opened 2024-10-08
Store state estimates into an HDF5 file

WardLT:
Adds ability to store each state estimation into an HDF5 file for later use.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -45,5 +45,7 @@ autoclass_content = 'both'
 intersphinx_mapping = {
     'python': ('https://docs.python.org/', None),
     'pandas': ('https://pandas.pydata.org/docs/', None),
-    'thevenin': ('https://rovi-org.github.io/thevenin', None)
+    'numpy': ('https://numpy.org/doc/stable/', None),
+    'thevenin': ('https://rovi-org.github.io/thevenin/', None),
+    'h5py': ('https://docs.h5py.org/en/stable/', None)
 }
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -14,6 +14,7 @@ Source Code: https://github.com/ROVI-org/auto-soh/
    system-models
    simulator
    estimators/index
+   toolkit
    source/modules
    extending
 
--- a/docs/source/interface.rst
+++ b/docs/source/interface.rst
@@ -5,3 +5,12 @@ Interface (``moirae.interface``)
    :members:
    :undoc-members:
    :show-inheritance:
+
+
+HDF5 Output (``m.interface.hdf5``)
+----------------------------------
+
+.. automodule:: moirae.interface.hdf5
+   :members:
+   :undoc-members:
+   :show-inheritance:
--- a/docs/toolkit.rst
+++ b/docs/toolkit.rst
@@ -35,10 +35,12 @@ and the estimator after running across all data.
 
 Access more state estimates by supplying a path at which to write an HDF5 file as the
 ``hdf5_output`` argument to ``run_online_estimate``.
-The interface will save the mean of the estimate for the state at each timestep,
-and a full resolution of the estimated state for the first timestep of each cycle.
+The interface will save the mean of the estimate for the state and outputs at each timestep,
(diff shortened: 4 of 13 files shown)


WardLT:
I'm almost certain we have a performance optimization for the writes possible. I believe we trigger a write of an entire chunk of file each time we ask for writing a single row. Need to see if there is already a buffering approach in h5py

WardLT:
There does not appear to be automatic buffering of writes, so I'll make a batching capability later. That will also involve removing the ability to write rows in a random order, which seems unnecessary anyway

WardLT:
Ok, I made a change which will simplify optimizing write performance later. Feel free to review the code as-is, Victor, as I'm going to start using this code on "real tests" to see how bad performance is.

WardLT:
Thanks, Ben and Claude!

victorventuri:
I have looked over the code, and maybe I lack the skills to fully understand all it is doing, so please correct me below: 

- the estimator state can be saved as `full` (in which case we use `pydantic`'s in-built `model_dump` method), or `mvn` (is this supposed to be 'Mean'N'(co)Variance'? if so, should the flag be `mnv` instead?), or just the mean, or not saving it at all. Is that correct? 
- does the same apply for the estimator prediction (the output, like the terminal voltage and its error bounds)?
- what maybe also isn't too clear to me is: are these states being saved at every time step, or at every cycle? I saw a mention of saving only the state at the first step of the new cycle, but ideally we would have the states at every time step, no?

I also went through the separate pdf you shared with me on slack; it is very nice! Would it be possible to also have a demo notebook showing how to save the states in full, summarized (mean and covariance), expected-only (just mean), at every time step?

WardLT:
Thanks for the feedback, Victor!

> if so, should the flag be mnv instead?), or just the mean, or not saving it at all. Is that correct?

I meant MVN to stand for multivariate normal. Shall I change it to "mean_cov" to cut down on confusion, especially if I add an option which stores only the mean and standard deviations?

> does the same apply for the estimator prediction (the output, like the terminal voltage and its error bounds)?

It would, if I remembered to store the outputs! (I'm going to add storing outputs too)

> are these states being saved at every time step, or at every cycle?

We store both every step and once per cycle, though you can turn either off. Is that useful? Would you prefer something else (e.g., an average over a cycle)?

> Would it be possible to also have a demo notebook?

Totally. I'll toss one together or add it to an existing example

victorventuri:
Thanks for the responses! Just adding here the follow-ups:

> Shall I change it to "mean_cov" to cut down on confusion, especially if I add an option which stores only the mean and standard deviations?

I think that would be best, so it's very clear we are only storing the first two moments of the distributions; if the user wants something more sophisticated, they can ask for `full`, or something simpler, just `mean`. 

> We store both every step and once per cycle, though you can turn either off. Is that useful? Would you prefer something else (e.g., an average over a cycle)?

I think I am starting to understand this a bit better, but just so I fully get it: we will have estimates saved for every time step, but the point at which they are removed from RAM and saved to disk can either happen at every time step or once per cycle. Is that the correct interpretation?

WardLT:
>I think I am starting to understand this a bit better, but just so I fully get it: we will have estimates saved for every time step, but the point at which they are removed from RAM and saved to disk can either happen at every time step or once per cycle. Is that the correct interpretation?

Pretty much. The minor difference is that they can be saved at disk _both_ at each step and once per cycle. Per-step and per-cycle are stored in different parts of the file

WardLT:
Alright, @victorventuri , I've made those improvements you suggested. That includes a demo notebook explaining how to use the HDF5 outputs.

https://github.com/ROVI-org/auto-soh/blob/hdf5_estimates/notebooks/interface-with-battery-toolkit.ipynb

Would you like to review again before merging?

victorventuri:
@WardLT just checked it out, looks very nice! I only have two further minor requests:
1. in the example, the only thing being estimated are the transient values. can we include some A-SOH estimate for completeness? it can be the simplest one, like `R0`, which only has one base value
2. can we demonstrate reading from the hdf5 file and reconstructing the original `MvRD` object in the case of saving the `full` state?

PR #96, opened 2024-10-27
Reworking batching interpolation

victorventuri:
Reworked the way interpolation works with batching, allowing the user (or user-application) to have greater flexibility in what to expect when both the SOC and the internal health variable are batched. 

By defaults, we try to match the SOC and internal batches, as well as remove superfluous batch dimensions of cardinality 1, as to successfully reproduce the behavior chosen earlier.


WardLT:
I'm likely to merge as-is when I'm back to work tomorrow. Docs and tests look good, and this code is internal enough that I'm not worried about changing behavior. I'll mostly look for performance issues.

Until then, two quick and mutually-exclusive suggestions:

1. Would a table be better than prose for explaining the output shapes?
2. Had you already considered requiring specific shapes and always returning a 3D array from the function? Adding and removing dimensions is very easy with NumPy arrays, so the tradeoff of "additional work to use" versus "difficulty to understand" is favorable.

victorventuri:
Thanks, Logan! Here's my followups:

> 1. Would a table be better than prose for explaining the output shapes?

I am not certain I understand. Are you referring to the documentation? If so, then maybe? I would have to look up how to do a proper table in markdown for rendering the docs, but otherwise, perhaps? I can try to write one and you can tell me what you think?

> 2. Had you already considered requiring specific shapes and always returning a 3D array from the function? Adding and removing dimensions is very easy with NumPy arrays, so the tradeoff of "additional work to use" versus "difficulty to understand" is favorable.

Yes, and perhaps simply always returning a 3D array is preferable. The reason I didn't do it was that this particular class is far-reaching inside `moirae`, and I tried to mimic previous behavior as to not break anything else. The balance in my mind is "how much work will it take to re-adjust everything that uses this" vs."how much does one need to understand the dimensionality concerns when using it" (which in my mind is a balance that is unclear)

Now, in terms of the 
>I'm likely to merge as-is when I'm back to work tomorrow.

my concern is the `test_interface.py` that is failing. I don't feel comfortable merging something that doesn't pass all the tests, but, at the same time, I don't believe I know how to fix it...

WardLT:
> Yes, and perhaps simply always returning a 3D array is preferable. 

If you think this is the better solution, then shall we implement it? Your IDE will help you track down all the calls to `get_values` and the unit tests are there to help catch the ones you miss it.

> my concern is the test_interface.py that is failing. 

I'm back from PTO and will tinker with that today. Ugh, conda

victorventuri:
> If you think this is the better solution, then shall we implement it?

I am not convinced it is, as I few there are very few cases in which we would not want to match the SOC batch with the health variable batch. However, I can try changing it to always being 3D, but I will say I am a bit wary of it; I don't think our tests are as thorough as coveralls would make us believe. It would also involve re-writing a significant amount of the tests I think

victorventuri:
Based on additional conversations with @WardLT, we opted to have the `get_value()` method of the `SOCInterpolatedHealth` class to always return a 3D array, so it's behavior is more predictable and easier to understand. This require a number of changes on the way the ECM classes operate, as well on a few tests. I am now just re-running a few of our demonstration notebooks to make sure they still perform as expected.

moirae/models/ecm/__init__.py
```diff
@@ -46,53 +46,66 @@ def update_transient_state(
...
+        current_slope = np.atleast_2d(0.0) if self.current_behavior == 'constant' \
+            else (current_kp1 - current_k) / delta_t  # shape (trans_batch, 1)
 
         # We will assume that all health parameters remain constant between time
         # steps, independent of temperature or SOC variations. The value used
         # will be the one at the previous SOC and temperature values.
 
         # Set Coulombic efficiency to 1. if discharging
-        coul_eff = 1 if current_k < 0 else asoh.ce
+        coul_eff = np.atleast_2d(1) if current_k < 0 else asoh.ce  # shape (ce_batch, 1)
```
WardLT, line 61: Not a required change, but a numpy suggestion: consider `np.ones((1, 1))` or `np.ones_line(asoh.ce)` instead.

moirae/models/ecm/__init__.py
```diff
@@ -141,28 +162,47 @@ def calculate_terminal_voltage(
...
+            i_rc = transient_state.i_rc  # shape (i_rc_batch, num_rc)
+            # For proper broadcasting, let's turn this into a shape of (num_rc, i_rc_batch, 1)
+            i_rc = i_rc.T[:, :, None]
+            V_drops = i_rc * rc_rs  # shape (rc.r_batch, num_rc, trans_batch, soc_dim=1)
+            V_drop_sum = np.sum(V_drops, axis=1)  # shape (rc.r_batch, trans_batch, soc_dim=1)
+            Vt = Vt + V_drop_sum  # shape (asoh_batch_size, trans_batch_size, 1)
 
         # Include hysteresis
-        Vt = Vt + transient_state.hyst
+        hyst = transient_state.hyst  # shape (hyst_batch, 1)
```
WardLT, line 198: A curiosity point: Are we finding examples where the ASOH and Transient batch sizes are different?

moirae/models/ecm/components.py
```diff
@@ -131,11 +131,11 @@ def get_value(self,
         """
         Returns values of OCV at given SOC(s) and temperature(s).
         """
-        ocv = enforce_dimensions(self.ocv_ref.get_value(soc=soc), 0)
```
WardLT, line 134: Lovely that we can get rid of these calls to the enforce_dimensions functions.

moirae/models/ecm/utils.py
```diff
@@ -43,40 +43,49 @@ def _interp_func(self) -> Callable:
         return func
 
     def get_value(self, soc: Union[Number, List, np.ndarray]) -> np.ndarray:
-        """Computes value(s) at given SOC(s)
+        """
+        Computes value(s) at given SOC(s).
+
+        This function always returns a 3D array, of shape `(internal_batch_size, soc_batch_size, soc_dim)`, where
```
WardLT, line 49: This is great. Much simpler compared to the previous behavior, and the implementation is shorter too! Nice to be able to delete code 👏🏼

WardLT (approved):
Nice work!

moirae/models/ecm/__init__.py
victorventuri: Not really. The only case so far was for the computation of the energy of a cell (see new PR)

moirae/models/ecm/components.py
victorventuri: Agreed!

moirae/models/ecm/utils.py
victorventuri: Thank you! Yeah, it is definitely much easier to debug :D

moirae/models/ecm/__init__.py
victorventuri: Is there a reason for choosing one or the other among the three options?

Issue #94, opened 2024-10-25
[FEATURE] ECM A-SOH objects should contain theoretical energy


victorventuri:
**Is your feature request related to a problem? Please describe.**
No, everything works as it should. However, in some instances (especially when generating data from a pre-determined power profile that needs to be scaled by a maximum energy term), it becomes necessary to know the value of the maximum theoretical energy stored in the cell easily, instead of having to compute it separately. 

**Describe the solution you'd like**
 It would be nice if I could do `energy = ECMASOH_obj.get_max_energy(temperature = temperature)`

**Describe alternatives you've considered**
The main alternative is having this calculation being done on the application side.

victorventuri:
Feature added by https://github.com/ROVI-org/auto-soh/pull/99

PR #99, opened 2024-11-01
Energy calculation in ECMASOH objects

victorventuri:
Adds feature requested in issue #94. The returned array is always 2D, with shape `(batch_size, 1)`


moirae/models/ecm/advancedSOH.py
```diff
@@ -27,6 +28,31 @@ class ECMASOH(HealthVariable):
     h0: HysteresisParameters = Field(default=HysteresisParameters(base_values=0.0),
                                      description='Hysteresis component')
 
+    def get_theoretical_energy(self,
+                               soc_limits: Tuple[float, float] = (0., 1.),
+                               temperature: Optional[float] = None) -> Union[float, np.ndarray]:
+        """
+        Function that computes the maximum theoretical energy of the cell in Wh.
+
+        By integrating the OCV over the SOC, we can calculate the amount of energy available in the cell, assuming no
```
WardLT, line 37: The tone here is a bit more academic than you find in most documentation. Let's switch from active "we" form to more concise, indicative form. 

"Computes cell energy by integrating Open-Circuit Voltage (OCV) over the supplied state of charge (SOC) ranges. Assumes no energy loss, such as from the resistive or hysteresis elements."

☝🏼 Simpler because there are fewer commas. Shorter because no active voice (No us, only code).

tests/models/ecm/test_asoh.py
```diff
@@ -33,3 +38,97 @@ def test_json(full_asoh):
...
+    ocv_ent = EntropicOCV(base_values=np.zeros(11))
+    ocv = OpenCircuitVoltage(ocv_ref=ocv_ref, ocv_ent=ocv_ent)
+    asoh = ECMASOH(q_t=qt, ce=1, ocv=ocv, r0=r0)
+
+    # With this simple example, we expect the energy to simply be the total discharge capacity multiplied by 1 V
+    expected = asoh.q_t.amp_hour
+    computed = asoh.get_theoretical_energy()
+    assert len(computed.shape) == 2, f'Computed energy is not 2D, but has shape {computed.shape}'
+    assert np.allclose(expected, computed), \
+        f'Mismatch in constant OCV ECMASOH energy calculation; expected {expected} but got {computed} Wh!'
```
WardLT, line 57: I like these messages from assert statements, but want to make sure you know they're not necessary. In fact, pytest will display the values of `expected` and `computed` in the failure message.

WardLT (approved):
I do like this. As discussed over Slack, these are my nit-picky-est comments.

moirae/models/ecm/advancedSOH.py
victorventuri: Changed the wording :D

tests/models/ecm/test_asoh.py
victorventuri: Got it, but I don't mind it too much. Makes it easier for me at least, since I run the tests from the terminal like a noob

PR #107, opened 2024-11-26
Adjust implementation to match Popovici 2017

WardLT:
Our present implementation differs from [at least one paper on dual UKFs](https://ntrs.nasa.gov/api/citations/20170005722/downloads/20170005722.pdf) in how updates to the ASOH parameters are informed.

Popovici uses only the effect of changing ASOH parameters on the outputs to determine how to update ASOH parameters, fixing the transient states to those predicted using the previous estimate of ASOH values.
See equation 13 below.

Our present implementation assesses the effect of changes in ASOH affecting the predicted transient states in addition to how changes in ASOH parameters affect the measurements.

image


WardLT:
After talking to @victorventuri , the current implementation follows exactly that in Plett's textbook on ECMs. So, let's not overwrite it. We'll still need to review other Dual Estimator implementations to see if my understanding of the Dual EKF (above) is an appropriate model for a Dual Estimator implemented as a secondary class.

WardLT:
Another point against the dual ekf is that it can only adjust ASOH parameters that are part of the output function, if @victorventuri and I understand it right

victorventuri:
That is not quite the case; below I try to explain it a bit more.

The Dual EKF should not have a problem with that, as it has explicit analytical derivatives of measurements with respect to hidden variables (that is, it requires something like $\frac{\partial V}{\partial Q}$), which is another reason the paper you linked can take this shortcut. 

The Dual UKF, however, relies on sampling to get the correlations between measurements and hidden (while the EKF assumes a linear relationship given by the first order derivative). Because of this, if we were to take that shortcut on the Dual UKF, we would not be able to adjust aSOH parameters that are independent of the measurement given the PHS.

WardLT:
Thanks, @victorventuri . I at least know enough now to know we're not merging this

PR #116, opened 2025-01-10
adding R0 extractor

npaulson:
deterministic instantaneous/series resistance extractor


WardLT:
Also, don't worry about the broken tests. As you may have guessed, it was broken for unrelated reasons.

PR #122, opened 2025-01-27
Add serialization and validation logic for SOC pinpoints

### LINKED ISSUE #120: [BUG] Error raised when recreating SOCInterpolatedHealth
**Describe the bug**
An `SOCInterpolatedHealth` variable may or may not have a `soc_pinpoints` field populated. In most instances, this field is populated only after a call to the `get_value()` method is made. Once this field is populated, however, the variable cannot be recreated with `model_dump_json()` and `model_validate_json()`: a pydantic `ValidationError` is raised. This also happens if we instantiate the variable with the SOC pinpoints

**To Reproduce**
```python
import numpy as np
from pydantic import ValidationError

from moirae.models.ecm.utils import SOCInterpolatedHealth

# Create a variable
variable = SOCInterpolatedHealth(base_values=np.arange(10))
assert variable.soc_pinpoints is None, 'SOC pinpoints created prematurely!'

# The usual stuff
varialb_str0 = variable.model_dump_json()
re_variable0 = SOCInterpolatedHealth.model_validate_json(varialb_str0)
assert np.allclose(variable.base_values, re_variable0.base_values), 'Wrong recreation of base values!'

# Now let's say we try to get a value, in which case, the soc_pinpoints are created
_ = variable.get_value(soc=0.5)
assert len(variable.soc_pinpoints) == variable.base_values.shape[1], 'Mismatch between SOC pinpoints and base values!'

# Now, try again
varialb_str1 = variable.model_dump_json()
try:
    re_variable1 = SOCInterpolatedHealth.model_validate_json(varialb_str1)
except ValidationError as error:
    print(error)

assert np.allclose(variable.base_values, re_variable0.base_values), 'Wrong recreation of base values!'

# Now, try pre-defining soc pinpoints
variable = SOCInterpolatedHealth(base_values=np.arange(10), soc_pinpoints=np.linspace(0, 1, 10))
assert len(variable.soc_pinpoints) == 10, 'SOC pinpoints defined incorrectly!'

# The usual stuff
varialb_str0 = variable.model_dump_json()
try:
    re_variable0 = SOCInterpolatedHealth.model_validate_json(varialb_str0)
except ValidationError as error:
    print(error)
assert np.allclose(variable.base_values, re_variable0.base_values), 'Wrong recreation of base values!'
```

**Expected behavior**
The object should be able to be stored to JSON and re-created faithfully.

**Desktop (please complete the following information):**
 - OS: macOS Sequoia 15.2
 - Version 0.0.1

**Additional context**
Tagging @WardLT for discussion.

WardLT:
Fixes #120 by adding proper handling for serialization and validation of a numpy array. Switches the pydantic serialization logic from the class-method style to the `Annotated` style we're using elsewhere in Moirae.

### DIFF
--- a/tests/models/ecm/test_ecm_utils.py
+++ b/tests/models/ecm/test_ecm_utils.py
@@ -85,3 +85,33 @@ def test_cube(cubic):
     assert np.isclose(values * values * values,
                       cubic.get_value(values),
                       atol=1e-12).all(), 'Wrong cubic interpolation!'
+
+
+def test_serialization():
+    # Create a variable
+    variable = SOCInterpolatedHealth(base_values=np.arange(10))
+    assert variable.soc_pinpoints is None, 'SOC pinpoints created prematurely!'
+
+    # Make sure the base_values serialize correctly
+    varialb_str0 = variable.model_dump_json()
+    re_variable0 = SOCInterpolatedHealth.model_validate_json(varialb_str0)
+    assert np.allclose(variable.base_values, re_variable0.base_values), 'Wrong recreation of base values!'
+    assert re_variable0.soc_pinpoints is None, 'SOC points where created on serialization'
+
+    # Now let's say we try to get a value, in which case, the soc_pinpoints are created
+    variable.get_value(soc=0.5)
+    assert len(variable.soc_pinpoints) == variable.base_values.shape[1], 'Mismatch between pinpoints and base values!'
+
+    # Ensure the base values get serialized correctly
+    varialb_str1 = variable.model_dump_json()
+    re_variable1 = SOCInterpolatedHealth.model_validate_json(varialb_str1)
+    assert np.allclose(variable.base_values, re_variable1.base_values), 'Wrong recreation of base values!'
+
+    # Now, try pre-defining soc pinpoints
+    variable = SOCInterpolatedHealth(base_values=np.arange(10), soc_pinpoints=np.linspace(0, 1, 10))
+    assert len(variable.soc_pinpoints) == 10, 'SOC pinpoints defined incorrectly!'
+
+    # The usual stuff
+    varialb_str0 = variable.model_dump_json()
+    re_variable0 = SOCInterpolatedHealth.model_validate_json(varialb_str0)
+    assert np.allclose(variable.base_values, re_variable0.base_values), 'Wrong recreation of base values!'
--- a/moirae/models/base.py
+++ b/moirae/models/base.py
@@ -56,6 +56,10 @@ ListParameter = Annotated[
 ]
 """Type annotation for parameters which can be any number of values"""
 
+NumpyType = Annotated[
+    np.ndarray, BeforeValidator(np.array), Field(), WrapSerializer(_encode_ndarray, when_used='json-unless-none')
+]
+
 
 # TODO (wardlt): Decide on what we call a parameter and a variable (or, rather, adopt @vventuri's terminology)
 # TODO (wardlt): Make an "expand names" function to turn the name of a subvariable to a list of updatable names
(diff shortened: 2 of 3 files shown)


tests/models/ecm/test_ecm_utils.py
```diff
@@ -85,3 +85,33 @@ def test_cube(cubic):
...
+def test_serialization():
+    # Create a variable
+    variable = SOCInterpolatedHealth(base_values=np.arange(10))
+    assert variable.soc_pinpoints is None, 'SOC pinpoints created prematurely!'
+
+    # Make sure the base_values serialize correctly
+    varialb_str0 = variable.model_dump_json()
+    re_variable0 = SOCInterpolatedHealth.model_validate_json(varialb_str0)
+    assert np.allclose(variable.base_values, re_variable0.base_values), 'Wrong recreation of base values!'
+    assert re_variable0.soc_pinpoints is None, 'SOC points where created on serialization'
```
victorventuri, line 99: I have a dumb question here: given the `BeforeValidator` is just casting a `np.array` on the argument, why does this work? As for as I know, `np.array(None) is None == False`

tests/models/ecm/test_ecm_utils.py
```diff
@@ -85,3 +85,33 @@ def test_cube(cubic):
...
+
+    # Now let's say we try to get a value, in which case, the soc_pinpoints are created
+    variable.get_value(soc=0.5)
+    assert len(variable.soc_pinpoints) == variable.base_values.shape[1], 'Mismatch between pinpoints and base values!'
+
```
victorventuri, line 104: Can we add a 
```python
assert np.allclose(re_variable0.get_value(0.5), variable.get_value(soc=0.5)), 'Wrong calculation of value!'
```
here for completeness?

tests/models/ecm/test_ecm_utils.py
```diff
@@ -85,3 +85,33 @@ def test_cube(cubic):
...
+    # Ensure the base values get serialized correctly
+    varialb_str1 = variable.model_dump_json()
+    re_variable1 = SOCInterpolatedHealth.model_validate_json(varialb_str1)
+    assert np.allclose(variable.base_values, re_variable1.base_values), 'Wrong recreation of base values!'
+
```
victorventuri, line 109: Can we add a 
```python
assert np.allclose(variable.soc_pinpoints, re_variable1.soc_pinpoints), 'Wrong recreation of SOC pinpoints!'
```
for completeness?

tests/models/ecm/test_ecm_utils.py
```diff
@@ -85,3 +85,33 @@ def test_cube(cubic):
...
+    # Now, try pre-defining soc pinpoints
+    variable = SOCInterpolatedHealth(base_values=np.arange(10), soc_pinpoints=np.linspace(0, 1, 10))
+    assert len(variable.soc_pinpoints) == 10, 'SOC pinpoints defined incorrectly!'
+
+    # The usual stuff
+    varialb_str0 = variable.model_dump_json()
+    re_variable0 = SOCInterpolatedHealth.model_validate_json(varialb_str0)
+    assert np.allclose(variable.base_values, re_variable0.base_values), 'Wrong recreation of base values!'
```
victorventuri, line 117: Similarly to above, can we have a 
```python
assert np.allclose(variable.soc_pinpoints, re_variable0.soc_pinpoints), 'Wrong recreationg of SOC pinpoints!'
```
here as well?

victorventuri (approved):

tests/models/ecm/test_ecm_utils.py
WardLT: My understanding is that the validator will only run if the value of field isn't `None`. I have it set up as `Optional[Annotated[..., BeforeValidator`, so Pydantic probably sees that the value already matches an allowed type (None) then skips any additional validation work.

tests/models/ecm/test_ecm_utils.py
WardLT: Done

tests/models/ecm/test_ecm_utils.py
WardLT: Done

PR #131, opened 2025-02-27
Use isotonic regression for OCV fitting

WardLT:
OCV must be monotonically increasing with SOC and
our current method does not enforce this

### DIFF
--- a/moirae/extractors/ecm.py
+++ b/moirae/extractors/ecm.py
@@ -5,6 +5,7 @@ import pandas as pd
 import matplotlib.pyplot as plt
 from sklearn.isotonic import IsotonicRegression
 from scipy.integrate import cumulative_trapezoid
+from scipy.interpolate import LSQUnivariateSpline
 from battdat.data import CellDataset
 from battdat.postprocess.integral import CapacityPerCycle
 
--- a/moirae/extractors/ecm.py
+++ b/moirae/extractors/ecm.py
@@ -1,12 +1,14 @@
 """Extraction algorithms which gather parameters of an ECM"""
+
 import numpy as np
 import pandas as pd
 import matplotlib.pyplot as plt
-from scipy.interpolate import LSQUnivariateSpline
+from sklearn.isotonic import IsotonicRegression
 from scipy.integrate import cumulative_trapezoid
-
 from battdat.data import CellDataset
 from battdat.postprocess.integral import CapacityPerCycle
+
+
 from moirae.extractors.base import BaseExtractor
 from moirae.models.ecm.components import ReferenceOCV, OpenCircuitVoltage, MaxTheoreticalCapacity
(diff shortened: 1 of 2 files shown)


moirae/extractors/ecm.py
```diff
@@ -1,12 +1,15 @@
 """Extraction algorithms which gather parameters of an ECM"""
+
 import numpy as np
 import pandas as pd
 import matplotlib.pyplot as plt
-from scipy.interpolate import LSQUnivariateSpline
+from sklearn.isotonic import IsotonicRegression
```
victorventuri, line 6: I have never used this form of regression before, so this may be a dumb question, but here goes nothing: does this capture monotonically strictly-increasing functions or monotonically non-decreasing functions? The OCV should be non-decreasing, but need not be monotonically strictly-increasing. For example, LFP cells usually have a wide region of flat OCV from SOCs of like 15 to 75% I think.

moirae/extractors/ecm.py
```diff
@@ -109,12 +112,9 @@ def interpolate_ocv(self, cycle: pd.DataFrame) -> np.ndarray:
...
         w /= w.sum()
 
-        # Evaluate the smoothing spline
-        t = self.soc_points[1:-1]
-        spline = LSQUnivariateSpline(
-            cycle['soc'].values, cycle['voltage'].values, w=w, t=t
-        )
-        return spline(self.soc_points)
+        # Fit then evaluate a monotonic function
+        model = IsotonicRegression(out_of_bounds='clip').fit(cycle['soc'], cycle['voltage'], sample_weight=w)
```
victorventuri, line 116: Here, it seems like we are fitting terminal voltage against SOC and setting that to be the OCV. This is only an OK approximation if the C-rate used to cycle the cell is very small, such that resistive terms don't play much of a role.  I think a C-rate check should be added to the `check_data` method. 

Alternatively, we can allow the `extract` method to be provided with the other ECM elements (R0, RC, hysteresis), which should have already been extracted from other sections of the RPT, and compute the OCV with knowledge of those.

victorventuri (approved):

moirae/extractors/ecm.py
WardLT: Good question. "Isotonic" in this sense means non-decreasing. Check out: https://scikit-learn.org/stable/modules/isotonic.html#isotonic

victorventuri: Nice, good to see this!

moirae/extractors/ecm.py
WardLT: Good point on the HPPC. Shall pause on that for a bit because I still need an OCV extractor that's useful w/o HPPC?

We can also implement multiple extractors for the same parameter, each with different algorithms/using different assumptions

PR #132, opened 2025-02-28
Add function for running model from data inside a dataset

WardLT:
A function I end up running often when testing moirae with data in battery-data-toolkit formats.


victorventuri (approved):
This looks great! Food for thought: should this be in the `__init__` of the interface, or part of a separate `utils.py` part (seeing as it is not a critical piece of the interface, but a nice-to-have functionality)?

WardLT:
Good question. I don't we yet have enough functions (just 2) in the interface module to start worry about spreading them out, even though this is more of a util function than something core

PR #135, opened 2025-03-05
Cache the SOC interpolation functions

WardLT:
Avoid re-fitting the SOC interpolation functions at each step.


moirae/models/ecm/components.py
```diff
@@ -28,7 +28,7 @@ def value(self, amp_seconds: float) -> None:
         self.base_values = amp_seconds / 3600.
 
     @property
-    def amp_hour(self) -> float:
+    def amp_hour(self) -> np.ndarray:
```
victorventuri, line 31: thank you, this had been bugging me for soooooo long and i always forgot to change it

victorventuri (approved):
I think this is fine, but I don't think we are yet at a point where we can take full advantage of this, as most of our interpolations are linear and come with close to no computational burden to be done on-the-fly/as needed.

moirae/models/ecm/components.py
WardLT: Knowing that message annoys you, I can do one better! I fixed the thousands of errors that come out of battdat's use of deprecated features in battery-data-toolkit: https://github.com/ROVI-org/battery-data-toolkit/pull/127

WardLT:
How about I add in some logic to skip caching for linear interpolation? I'd like to keep this logic in here so that we can use it later if we move to cubic splines or whatever

moirae/models/ecm/components.py
victorventuri: Fantastic :heart_eyes:! are they merged to main on the toolkit side?

victorventuri:
Yeah, that's a good idea! This way, we know that everything we have already done should be working exactly the same, but we have now access to additional functionality as well!

PR #139, opened 2025-03-19
Polish the r0 extractor

WardLT:
Make a few changes to the R0Extractor:

- [x] Remove plotting code, which becomes tedious to maintain (e.g., continually adding arugments to change style)
- [x] Avoid persisting results from the call to `extract`. We don't describe the Extractors as stateful in the documentation
- [x] Use battery-data-toolkit's logic for computing SOC


moirae/extractors/ecm.py
```diff
@@ -210,15 +206,18 @@ def interpolate_r0(self, cycle: pd.DataFrame) -> np.ndarray:
...
+                       (cycle['cycle_capacity'].max() - cycle['cycle_capacity'].min())
 
         # calculate instantanous resistance at all points
         cycle['r0_inst'] = np.abs(
             cycle['voltage'].diff() /
             cycle['current'].diff())
 
         # calculate key quantities to filter R0_inst
-        Inorm = cycle['capacity'].max()
+        Inorm = cycle['cycle_capacity'].max() - cycle['cycle_capacity']
```
npaulson, line 220: I don't think this is the correct thing to do. We should keep the original as we want it to be a scalar quantity that we use to normalize the change in current.

moirae/extractors/ecm.py
```diff
@@ -210,15 +206,18 @@ def interpolate_r0(self, cycle: pd.DataFrame) -> np.ndarray:
         """
         # calculate soc throughout cycle
         cycle = cycle.copy(deep=False)
-        cycle = self.soc_calc(cycle)
+        if 'cycle_capacity' not in cycle.columns:
+            StateOfCharge().enhance(cycle)
```
npaulson, line 210: I have concerns about the methods used to calculate the change in capacity and energy within a given cycle within the Battery Data Toolkit. I will make an issue over there.

WardLT: Absolutely. We do need to polish that algorithm up.

WardLT: Are you cool with using the battery-data-toolkit implementation until we improve it?

npaulson (commented):
I think the change in line 221 of moirae/extractors/ecm.py should be reverted. I am also concerned about the battery data toolkit capacity calculations, but I created an issue in that package. Otherwise, I am okay removing plotting and the redundant capacity/soc calculations.

WardLT:
Yea, Line 221 was definitely wrong. I did fix it in a newer commit but neglected to reply to your comment: 

It now gets the capacity using the difference between the maximum and minimum cycle-capacity
```
Inorm = cycle['cycle_capacity'].max() - cycle['cycle_capacity'].min()
```

PR #141, opened 2025-03-25
Add ability to include priors in loss function

WardLT:
Adds ability to add a "constraint" to the solution off offline estimators based on the prior probability distribution for parameters.

Changes:
- [x] Refactoring to make adding more loss functions simpler
- [x] Introduce a "PriorLogLoss" objective function
- [x] Adding ability to add loss functions together
- [x] Expand documentation for offline estimators


moirae/estimators/offline/loss/__init__.py
```diff
@@ -0,0 +1,117 @@
...
+           asoh_priors={},
+           cell_model=ecm_model,
+           asoh=init_asoh,
+           transient_state=int_state,
+           observations=timeseries_dataset
+       )
+
+    """
+
+    def __init__(self,
```
victorventuri, line 84: Does this class need to redefine the `get_x0` method to be based on the prior distributions, rather than the provided aSOH and transient objects? (that is, should the `get_x0` return be stochastic or deterministic?)

moirae/estimators/offline/loss/__init__.py
```diff
@@ -0,0 +1,117 @@
...
+
+        # Compute the mean-squared-error for each member of the batch
+        squared_error = np.power(pred_y - true_y, 2)
+        return np.mean(squared_error, axis=(0, 2))  # Average over steps and outputs
+
+
+class PriorLoss(BaseLoss):
+    """Compute the negative log-probability of parameter values from a prior distribution
+
+    Supply priors as a scipy :class:`~scipy.stats.rv_continuous` distribution
```
victorventuri, line 62: For now, I think it's fine to rely on the implementations available elsewhere for this, but I am not a fan of being constrained by them. Given we already have the formalism to compute log-likelihood, I would prefer we use the classes we define, and which we have control over. Obviously, that would require some additional logic, but could also cut down on some code in here. 

Another point to be made here is that, in some cases, we may wish to represent a multivariate RV in a way that is no available in `scipy` (for example, if we do at some point implement a particle filter, the RV is represented by discrete particles, and is not quite considered a continuous distribution)

moirae/estimators/offline/loss/__init__.py
```diff
@@ -0,0 +1,117 @@
...
+
+            true_y[i + 1, :] = new_out.to_numpy()
+            pred_y[i + 1, :] = pred_out.to_numpy()
+
+        # Compute the mean-squared-error for each member of the batch
+        squared_error = np.power(pred_y - true_y, 2)
+        return np.mean(squared_error, axis=(0, 2))  # Average over steps and outputs
+
+
+class PriorLoss(BaseLoss):
```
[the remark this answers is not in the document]
WardLT: I'm not sure yet, but would probably add another class for that because both the logic for defining it and for computing the probabilities would be different.

tests/estimators/offline/test_objectives.py
```diff
@@ -18,13 +19,70 @@ def test_mse_loss(simple_rint, timeseries_dataset, state_only):
...
-
     # Run the evaluation
+    x0 = loss.get_x0()[None, :]
     y = loss(x0)
     assert y.shape == (1,)
     assert np.isclose(y, 0.)
+
+
+@mark.parametrize('state_only', [True, False])
+def test_prior_loss(simple_rint, timeseries_dataset, state_only):
```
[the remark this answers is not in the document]
WardLT: As above, maybe we make that another feature? The priors I have an immediate need to set have no covariance between parameters.

moirae/estimators/offline/loss/__init__.py
WardLT: Hmm 🤔 That's a cool idea, but has some pitfalls I'd hope to avoid. It's not immediately clear which point from a prior you want to start with. Not all distributions have a well-defined "mode" (e.g., uniform), the mean of some (e.g., dual-peaked normal distributions), and drawing a random sample could be problematic. 

Since there's a not a clear, universal answer. I'd vote for keeping the current behavior: go off of whatever the user provided in the initial ASOH/state.

moirae/estimators/offline/loss/__init__.py
WardLT: I thought much about using `scipy` distributions, given that we have some distributions objects in moirae already, and believe it's better here.

> but I am not a fan of being constrained by them

You can always implement new multivariate distributions which follow the interface defined by SciPy. 

> Given we already have the formalism to compute log-likelihood, I would prefer we use the classes we define, and which we have control over. 

The only thing we cannot do with using SciPy as a base is change the behavior of SciPy functions.

> Another point to be made here is that, in some cases, we may wish to represent a multivariate RV in a way that is no available in scipy (for example, if we do at some point implement a particle filter, the RV is represented by discrete particles, and is not quite considered a continuous distribution)

That's an issue which isn't relevant here. If it isn't a continuous distribution, then we cannot use it to define priors.

Overall, it may even be worth converting our multivariate classes to build atop the SciPy's interface rather than defining or own.

tests/estimators/offline/test_objectives.py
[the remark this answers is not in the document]
WardLT: We can represent those, so long as there are no covariances between the variables:

E.g., both `q_t.base_values` and `r0.base_values`, provide both as inputs

```python
  prior_loss = PriorLoss(
      transient_priors={},
      asoh_priors={
      'q_t.base_values': ...,
      'r0.base_values': ...,
},
      cell_model=ecm_model,
      asoh=rint_asoh,
      transient_state=rint_state,
      observations=timeseries_dataset
  )
 ```
 
 You can also make a normal distribution of a list of variables using the same `norm` object
 
 ```
 >>> from scipy.stats import norm
>>> norm(loc=[0.1, 0.2], scale=[0.1, 0.2])
<scipy.stats._distn_infrastructure.rv_continuous_frozen object at 0x71f070ec48e0>
>>> y = norm(loc=[0.1, 0.2], scale=[0.1, 0.2])
>>> y.stats()
(array([0.1, 0.2]), array([0.01, 0.04]))
>>> 
```

PR #145, opened 2025-04-01
Rc extractor

### LINKED ISSUE #101: Quick initial estimates of ECM parameters
Be able to estimate OCV and resistive elements in a single pass, used before we go into a full optimization tool

npaulson:
addition of RC extractor, tests, documentation, and docs notebook content

Fixes #101


moirae/extractors/ecm.py
```diff
@@ -256,3 +257,277 @@ def extract(self, dataset: CellDataset) -> Resistance:
...
+        grp = cycle.groupby('step_index')
+        step_data_prev = None
+
+        rests = []
+        socs = []
+        for step, step_data in grp:
+
+            Iavg = step_data['current'].mean()
+            Istd = step_data['current'].std()
+            if np.abs(Iavg) > 0.1 or Istd > 0.1:
```
victorventuri, line 409: I think these thresholds should be
- a function of the capacity (for example, in some CAMP cells, a current of 0.1 A is almost C/3); say, for example, an average current that corresponds to at max a C-rate of C/20 
- specified by the user when initializing the function (though they can have default values)

moirae/extractors/ecm.py
```diff
@@ -256,3 +257,277 @@ def extract(self, dataset: CellDataset) -> Resistance:
...
+    """
+
+    soc_points: np.ndarray
+    """State of charge points at which to estimate the resistance"""
+    soc_requirement: float
+    """Require that dataset samples at least this fraction of the capacity"""
+    n_rc: int
+    """Number of RC couples in the ECM"""
+    min_rest: float
+    """Minimum required rest duration in seconds"""
```
victorventuri, line 292: Given we append the previous two entries of the `step_data_prev` to the present `step_data` (like 418), shouldn't we also need to know what the value of R0 is? In most instances, the first drop (or increase) in voltage when the current suddenly goes to zero is more due to R0 than RC elements, so shouldn't we first make sure that contributions is adequately dealt with?

tests/extractors/test_ecm_rc.py
```diff
@@ -0,0 +1,46 @@
...
+    assert np.allclose(rc_points[0][0], expected_rc[0], rtol=5e-2), Rdiff
+
+    Cdiff = f'C max diff: {np.abs(rc_points[0][1] - expected_rc[1]).max():.2e}'
+    assert np.allclose(rc_points[0][1], expected_rc[1], rtol=5e-2), Cdiff
+
+
+def test_full(rc_dataset, rc_extractor):
+    rcs = rc_extractor.extract(rc_dataset)
+    assert np.isclose(rcs[0].get_value(0.)[0], unrealistic_fake_rc(0.)[0], rtol=5e-2)
+    assert np.isclose(rcs[0].get_value(0.)[1], unrealistic_fake_rc(0.)[1], rtol=5e-2)
```
victorventuri, line 46: Could we also create a test that checks for more than one RC pair?

moirae/extractors/ecm.py
```diff
@@ -256,3 +257,277 @@ def extract(self, dataset: CellDataset) -> Resistance:
...
+        3. Fit :attr:`n_rc` exponential models to the IV(t) data
+        4. The scale and power parameters are converted to the RC parameters
+        5. Fit a 1-D smoothing cubic spline for voltage as a function of SOC,
+           placing knots at :attr:`soc_points`.
+        6. Evaluate the spline at SOC points requested by the user,
+           return as a :class:`~moirae.models.ecm.components.Resistance` object
+           using the :attr:`interpolation_style` type of spline.
+
+    Args:
+        soc_points: SOC points at which to extract R0 or (``int``) number of
```
WardLT, line 281: Could you document the other arguments to `__init__` here as well?

moirae/extractors/ecm.py
```diff
@@ -256,3 +257,277 @@ def extract(self, dataset: CellDataset) -> Resistance:
...
+        As = params[0::2]
+        Ts = params[1::2]
+        for i_rc in np.arange(self.n_rc-1)[::-1]:
+            Ts[i_rc] *= Ts[i_rc+1]
+
+        t = np.array(t_fitseg) - np.array(t_fitseg)[0]
+        t = t[..., None]
+
+        res = cycle_data['voltage'][indx_fitseg[-1]]
+        res *= np.ones((t.size, npset))
```
WardLT, line 489: Why multiply by ones here? Is it to reshape or tile the array?

moirae/extractors/ecm.py
```diff
@@ -256,3 +257,277 @@ def extract(self, dataset: CellDataset) -> Resistance:
...
+        ref = cycle_data['voltage'][indx_fitseg]
+        ref = np.array(ref)[:, None]
+
+        err = np.sum((ref-res)**2, axis=0)
+
+        err[np.isnan(err)] = np.inf
+
+        return np.sum(err)
+
+    def extract(self, dataset: CellDataset) -> Resistance:
```
WardLT, line 508: Could you fix the return type for this argument? It's a tuple of RCComponents, right?

WardLT (changes requested):
Just a few minor documentation-related changes and one question about a numpy operation I'm confused about. Everything else looks fine to me.

victorventuri (commented):
Great work! I've added a few comments on the relevant python scripts, but I will add some more here because commenting the notebook is kind of annoying on my side (idk how to comment on the properly rendered ipynb, only on the "raw" version). I guess I had not seen the demo_extractor notebooks before, so some of the comments are not directly related only to the RC extraction.

- I am in favor of using real data (e.g. CAMP) for demo purposes, but I think that using synthetic data is also important, so that we can show the extractors actually work (that is, they attain the ground truth, or very close to it, as the last few plots show). Similarly, when using real data, I think it would be valuable to, at the end of the full extraction, plot the real measured voltage vs. the voltage that would be generated by an ECM with the extracted parameters
- I made a similar comment on the RC extractor code, but I think knowledge of R0 is also necessary for the OCV extraction. I know the hope is that IR contributions at low C-rate should be very small compared to OCV ones, but it's always safer to include them.
- For the synthetic data used in the notebook, where did it come from? How would one (not us) know the ground truth values? 
- Are we set on requiring that the data be stored in the toolkit format, or just formatted accordingly? That is, do we want to enforce that the raw data be in `data.tables['raw_data']`, or should we instead make it so that arguments to the functions are `pd.DataFrame`s that follow the format from the toolkit?

WardLT:
> Are we set on requiring that the data be stored in the toolkit format, or just formatted accordingly? That is, do we want to enforce that the raw data be in data.tables['raw_data'], or should we instead make it so that arguments to the functions are pd.DataFrames that follow the format from the toolkit?

Let's punt on this question until we have a reason to change.

WardLT:
> I am in favor of using real data (e.g. CAMP) for demo purposes, but I think that using synthetic data is also important, so that we can show the extractors actually work (that is, they attain the ground truth, or very close to it, as the last few plots show). Similarly, when using real data, I think it would be valuable to, at the end of the full extraction, plot the real measured voltage vs. the voltage that would be generated by an ECM with the extracted parameters

Let's also punt on this until we get comfortable enough with the implementation. I'd like this PR to introduce a v0 that we can tweak and improve later

moirae/extractors/ecm.py
```diff
@@ -256,3 +257,277 @@ def extract(self, dataset: CellDataset) -> Resistance:
...
+            cycle: Dataset containing the time series measurement
+        Returns:
+            A list of relevant rest segments with associated soc, rest indices
+            rest times, and the state and average current of the previous step
+        """
+
+        cycle = cycle.copy(deep=False)
+        cycle['soc'] = cycle['cycle_capacity'] / self.capacity  # Ensure data are [0, 1)
+
+        grp = cycle.groupby('step_index')
```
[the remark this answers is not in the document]
WardLT: Can you raise an Issue in battery-data-toolkit to revisit our step detection logic. It's super old, predating the battery-data-toolkit even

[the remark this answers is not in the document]
npaulson: yes, I think it is important that we have good step and sub-step identification. It makes operations on the battery data much easier.

moirae/extractors/ecm.py
npaulson: Sure! I'm a bit confused with the difference between documenting the args in the docstring versus just below it. Something to do with pydantic?

moirae/extractors/ecm.py
npaulson: Yes, it is to tile the array for the purpose of vectorization, in case the optimizer requests the errors for multiple parameter sets. I believe differential evolution can take advantage of this.

moirae/extractors/ecm.py
npaulson: done!

moirae/extractors/ecm.py
```diff
@@ -256,3 +257,277 @@ def extract(self, dataset: CellDataset) -> Resistance:
...
+           placing knots at :attr:`soc_points`.
+        6. Evaluate the spline at SOC points requested by the user,
+           return as a :class:`~moirae.models.ecm.components.Resistance` object
+           using the :attr:`interpolation_style` type of spline.
+
+    Args:
+        soc_points: SOC points at which to extract R0 or (``int``) number of
+        grid points.
+    """
+
```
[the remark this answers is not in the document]
npaulson: Done!

moirae/extractors/ecm.py
npaulson: The purpose of appending the previous two entries is to ensure we get the entire rest period and to detect the state of the previous step (charge or discharge). We are not using any of the data from the charge/discharge step in the fit.

moirae/extractors/ecm.py
WardLT: Short answer: Yes, pydantic.

Longer answer: Both are fine and equivalent to the tool which renders our document. I prefer putting the documentation for the arguments in the docstring of the class (rather than for the `__init__` function) because pydantic auto-generates the `__init__` function. So, I try to use the class docstring so I can be consistent across all classes.

moirae/extractors/ecm.py
WardLT: Excellent. The way you do it here works great. There is a [`np.tile` function](https://numpy.org/doc/stable/reference/generated/numpy.tile.html) if you want to tiling the the same way

moirae/extractors/ecm.py
```diff
@@ -256,3 +257,277 @@ def extract(self, dataset: CellDataset) -> Resistance:
...
+        Args:
+            cycle: Dataset containing the time series measurement
+        Returns:
+            A list of relevant rest segments with associated soc, rest indices
+            rest times, and the state and average current of the previous step
+        """
+
+        cycle = cycle.copy(deep=False)
+        cycle['soc'] = cycle['cycle_capacity'] / self.capacity  # Ensure data are [0, 1)
```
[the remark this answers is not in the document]
npaulson: Good point - instead, I have added the relevant part of the self.check_data to _extract_rests because check_data calls _extract_rests, leading to a recursion. The purpose of having check_data do this is to ensure that there are sufficient relevant rest periods.

moirae/extractors/ecm.py
npaulson: Good point! I'll play arounds and find something reasonable.

moirae/extractors/ecm.py
victorventuri: I see. In this case, I'm a little confused. Don't you use the `Iprev` to compute the resistive term(s)?

tests/extractors/test_ecm_rc.py
npaulson: It was a bit of a task, but done!

moirae/extractors/ecm.py
npaulson: Yes, R = A/Iprev. Can you tell me what the source of the confusion is?

victorventuri: So, we fit a standard exponential decay of the form $A*\exp(-t/\tau)$, where $\tau=R*C$ and $A$ corresponds to the overpotential across the RC pair at time $t=0$. When we set $R=A/I_{prev}$, the assumption is that the total current flowing through the cell at the previous step is also flowing through the resistive component of the RC pair, and none of it is being used to charge/discharge the capacitive element. While this is true a good portion of the time, it fails in cases where that current was only applied for a short amount of time (shorter than $\tau$), which is usually the case in pulses (after a 5s pulse, for example, it is possible for the current flowing through the resistive element in the RC to not match the total cell current). 

A different way to get to what that overpotential should be is by looking at the voltage right before the rest, and by subtracting from it the Ohmic $R_0 * I_{prev}$ term, so that you make no assumptions as to what the current through the resistive element should be.

moirae/extractors/ecm.py
npaulson: I switched to using np.tile, thanks!

moirae/extractors/ecm.py
npaulson: Okay, let's chat about this a bit more next week

victorventuri: Noah and I chatted about this earlier. It doesn't have to be fixed for this PR, as the assumption is valid 99.9% of the time. But it is an issue that we should work on later, to make sure we don't fit the rest in-between pulses, for example.

WardLT:
LMK when you're ready to merge this, @npaulson

PR #146, opened 2025-04-08
Raise error if parameter does not exist

### LINKED ISSUE #39: Ensure that `mark_updatable` yields error if parameter does not exist
I just tried to mark `r_int.base_value` as updatable when `r_int.base_value*s*` is the actual name of parameters.

WardLT:
Fixes #39


victorventuri (approved):
Seems fine by me! Nice catch!

PR #147, opened 2025-04-18
Reduce the number of copies made when running UKF

WardLT:
It is expensive to make copies of HealthVariables and GeneralContainers. This PR changes the CellModelWrapper so it it modifies its internal copies of the variables rather than copying them at each timestep.

### DIFF
--- a/moirae/estimators/online/filters/kalman/unscented.py
+++ b/moirae/estimators/online/filters/kalman/unscented.py
@@ -88,6 +88,7 @@ class UnscentedKalmanFilter(BaseFilter):
         covariance_process_noise: covariance of process noise (default = 1.0e-8 * identity)
         covariance_sensor_noise: covariance of sensor noise as (default = 1.0e-8 * identity)
     """
+
     def __init__(self,
                  model: ModelWrapper,
                  initial_hidden: MultivariateGaussian,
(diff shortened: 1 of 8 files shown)


moirae/estimators/online/filters/kalman/unscented.py
```diff
@@ -271,25 +245,6 @@ def _assemble_unscented_estimate(self, samples: np.ndarray) -> Dict:
                                                                   cov_weights=self.cov_weights)
         return unscented_info
 
-    def _predict_outputs(self,
```
victorventuri, line 274: I understand removing these for streamlining the code. I had included them here as they could serve as a template if we were to implement other types of filter, in which some additional manipulation happens (such as re-weighing of some samples, or something of that nature), since most Kalman (and Bayesian) filters go through a similar estimate-correct type of stepping, but the inner working of each type could be a bit different. If having these additional calls is compromising the efficiency of the runs, then we should certainly remove them.

victorventuri (approved):
Looks good!

moirae/estimators/online/filters/kalman/unscented.py
WardLT: Ah, pardon me being too aggressive in pruning things. I thought these functions were from before we introduced the wrapper

I agree it's a good idea to have them as a route to ensure subclassing's eventually easier. I'll add them back in.

PR #148, opened 2025-04-21
Use Thevenin's predictor interface

WardLT:
The Prediction interface of Thevenin is dramatically faster than the Model interface for our step-by-step solutions. See Thevenin's docs, https://thevenin.readthedocs.io/latest/examples/prediction_interface.html

### DIFF
--- a/moirae/models/thevenin/__init__.py
+++ b/moirae/models/thevenin/__init__.py
@@ -3,10 +3,10 @@ import numpy as np
 from functools import partial
 from typing import Iterator
 
-from thevenin import Model, Experiment
+from thevenin import Prediction, TransientState
 
 from ..base import CellModel, InputQuantities, OutputQuantities
-from .state import TheveninASOH, TheveninTransient
+from moirae.models.thevenin.state import TheveninASOH, TheveninTransient
 from .ins_outs import TheveninInput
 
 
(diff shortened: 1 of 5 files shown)


tests/models/test_thevenin.py
```diff
@@ -178,19 +178,19 @@ def test_estimator():
...
     )
-    assert est.state.get_covariance().shape == (4, 4)
+    assert est.state.get_covariance().shape == (5, 5)
 
     est_state, est_outputs = est.step(new_inputs, expected_v)
     assert np.isclose(est_outputs.get_mean(), expected_v.terminal_voltage)
     est_mean = est_state.get_mean()
-    assert np.allclose(expected_state.to_numpy(), est_mean[:2])  # SOC, T
-    assert np.allclose(asoh.r[0].t_coeffs, est_mean[2:])  # Temp dependence of R
+    assert np.allclose(expected_state.to_numpy(), est_mean[:3], rtol=1e-4)  # SOC, T, hyst
```
victorventuri, line 192: To me, this doesn't feel like a proper test of the estimator, as it only makes sure one step does not deviate significantly from the ground truth at which the parameters were initialized. Given we do not control `thevenin`, I find it even more important to make sure the estimator that uses it converges to relevant statistics when initialized not exactly at the ground truth. I've tried to do that [here](https://github.com/ROVI-org/auto-soh/blob/main/tests/estimators/online/test_joint_ukf.py#L99) when testing the joint estimator. This will make this test slower, but, if we need to, we can mark it as slow and let the weekly test-runs raise issues if something ever comes up. (it shouldn't be too bad though, seeing as `thevenin` has a much better simulation capability than pure `moirae`)

victorventuri (changes requested):
I am not as familiar with the `Prediction` interface for `thevenin`, so it's not as easy for me to fully grasp how this is working in a short period of time. Before merging, however, I think we need to make sure the estimator that is based on `thevenin` converges when initialized away from the ground truth over a certain number of cycles.

moirae/models/thevenin/__init__.py
```diff
@@ -21,51 +21,82 @@ class TheveninModel(CellModel):
...
+            transient: Transient state
+
+        Returns:
+            - Dictionary of parameters to applied to Predictor
+            - Transient state of the battery
+        """
+        # The value of each member of the transient or ASOH is a 2D array with the first dimension either 1
+        #  or batch_size. The % signs below are a short syntax for either using the same value for all batches
+        #  (anything mod 1 is 0) or the appropriate member of the batch
+        params = {'num_RC_pairs': asoh.num_rc_elements, 'isothermal': self.isothermal, 'soc0': 0.}
```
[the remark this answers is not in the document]
WardLT: I don't believe `soc0` is actually used, but updated this line to use the current SOC regardless

tests/models/test_thevenin.py
WardLT: I agree with your desire to further test the joint estimator, but I'd push back on doing it here. I only intend this test to ensure that the interface works plausibly, and rely on tests for the joint estimator to assure correctly.

WardLT: >  Given we do not control thevenin, I find it even more important to make sure the estimator that uses it converges to relevant statistics when initialized not exactly at the ground truth.

Also, we're just going to rely on thevenin being correct. We cannot control everything.

Would you prefer a test against an analytical solution of an ECM rather than assuring that results are self consistent with calling thevenin a different way?

WardLT:
>Before merging, however, I think we need to make sure the estimator that is based on thevenin converges when initialized away from the ground truth over a certain number of cycles.

I do not think testing convergence here is appropriate for a handful of reasons:
- The test for thevenin should be limited to ensuring that thevenin works. That test around convergence feels like it's testing the reliability of the joint estimator, which should be done elsewhere
- I don't think the convergence tests we have now are that great of unit tests anyway. Many mistakes can be masked or at least convoluted if we're only testing the answer after thousands of iteration of an algorithm.

I'd be open to making a unit test where I compare the solution of a step of Joint Estimation against one done "by hand" (i.e., outside of Moirae), but otherwise believe these tests to be sufficient.

tests/models/test_thevenin.py
victorventuri: My primary concerns is that the way `thevenin` handles transient and health variables is different from what we do in `moirae` (at least, it was so in the old `Model` class from `thevenin`). For example, updating the health variables required some additional re-preparation of the model (there is a `pre()` method that is called every time the model was initialized, and then again every time the parameters are updated)

victorventuri: > I only intend this test to ensure that the interface between Thevenin and the estimator works plausibly, and rely on tests for the joint estimator to assure the estimator is correct.

I am confident the estimator is correct (we have tests for that), and certain `thevenin` is also correct (after using it and seeing Corey's numerous tests as well). However, due to the intricacies of how `thevenin` deals with transient and health variables, I fear a single step of the estimator is insufficient to demonstrate the connection between the codebases is robust. Perhaps the demonstration against an analytical solution of the ECM directly from `thevenin` would help put my mind at ease...

WardLT: What about 3-4 steps? If we're worried about the re-preparation. Let's shoot for something minimal

victorventuri:
> I don't think the convergence tests we have now are that great of unit tests anyway. Many mistakes can be masked or at least convoluted if we're only testing the answer after thousands of iteration of an algorithm.

I don't fully agree with this point. To be clear, I am not opposed to testing a single step and other small functionalities, quite on the contrary, I am all for those. However, the main point of the estimator is to be able to converge to an (user-)uncertain ground truth, and to track it thereafter in an online fashion. That's why I found it critical to testing this large-scaled functionality of both the joint and dual estimators we have, even if it may mask other mistakes (which is why I also believe the smaller functionalities, like a single `step` call, should also be tested, and it was my oversight in not doing so in the currently existing tests). 

> I'd be open to making a unit test where I compare the solution of a step of Joint Estimation against one done "by hand" (i.e., outside of Moirae), but otherwise believe these tests to be sufficient.

I'm 100% in favor of this! If you think these tests are sufficient, I trust you, and am good to merge.

victorventuri:
(accidentally closed the PR, but just reopened it, my bad)

WardLT:
Sounds like we need a few little tiers of tests: 1 step, 4 steps, 10k steps. :)

I'll piece out the middle-sized number of steps with Thevenin tomorrow and see where that gets us.

victorventuri:
Sounds good!

WardLT:
Alright, I've added a test where we run 500 steps (pardon the commit message stating 1000s) and show that the SOC is within 5% of correct even after perturbing the resistance from the true value.

I also added a test that highlights the batching.

Does that alleviate your concerns about the interface between moirae and thevenin?

victorventuri:
Thanks @WardLT ! 

> Does that alleviate your concerns about the interface between moirae and thevenin?

For the most part, yes! I'm for merging it as is and dealing with minor things later if they come up as issues.

PR #149, opened 2025-04-22
Revise detecting and iterating over health parameters

WardLT:
Fix a few issues:
- [x] `self.model_fields` is deprecated in Pydantic
- [x] Mark health parameters explicitly rather than assume all numpy arrays
- [x] Reduce the amount we test types of objects

The change is to add two properties to `HealthVariable`, `all_fields` and `all_fields_with_types`, that hold the logic to determine which fields (attributes) of the HealthVariable map to health parameters. 

How parameters are detected is slightly changed: it looks for numpy arrays tagged with "moirae_parameter" via Annotated tagging and HealthVariables (or collections of them) are found by explicitly testing whether a type is a `Tuple[HealthParameter]` or `Dict[str, HealthParameter]` via Python's issubclass logic.

### DIFF
--- a/moirae/models/base.py
+++ b/moirae/models/base.py
@@ -50,7 +50,7 @@ def _encode_ndarray(value: np.ndarray, handler) -> List:
 
 ScalarParameter = Annotated[
     np.ndarray,
-    'moirae_parameter',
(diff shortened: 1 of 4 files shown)


victorventuri:
Looks good to me!

victorventuri (approved):

PR #155, opened 2025-04-30
Allow using Thevenin with Offline Estimation

WardLT:
Make the class used to makes inputs/outputs from a battdat dataframe in the MSELoss configurable.

Also promoted "temperature" from an attribute that is specific in Moirae-ECM to one used in all models. We may revisit later. I'm punting for now because we don't have any models which don't use temperature.

### DIFF
--- a/moirae/models/base.py
+++ b/moirae/models/base.py
@@ -596,6 +596,7 @@ class InputQuantities(GeneralContainer):
     """
     time: ScalarParameter = Field(default=0., description='Timestamp(s) of inputs. Units: s')
     current: ScalarParameter = Field(default=0., description='Current applied to the storage system. Units: A')
+    temperature: Optional[ScalarParameter] = Field(None, description='Temperature reading(s). Units: °C')
 
 
 class OutputQuantities(GeneralContainer):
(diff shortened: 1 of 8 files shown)


moirae/models/base.py
```diff
@@ -596,6 +596,7 @@ class InputQuantities(GeneralContainer):
     """
     time: ScalarParameter = Field(default=0., description='Timestamp(s) of inputs. Units: s')
     current: ScalarParameter = Field(default=0., description='Current applied to the storage system. Units: A')
+    temperature: Optional[ScalarParameter] = Field(None, description='Temperature reading(s). Units: °C')
```
victorventuri, line 599: Well, I thought this was already here haha Nice one! we might want to think more where temperature should go, since, as it turns out, if is is measured, it is technically also an `OutputQuantity`...

victorventuri (approved):
You know more about the offline estimations that I do, so I am not much help on those changes, but happy to see we can use Thevenin for that too!

moirae/models/base.py
WardLT: Yea, temperature's a sticky point. I'll open an Issue so we can recall this discussion later

PR #161, opened 2025-05-06
Remove CellDataset from moirae

WardLT:
We've removed specialized dataset subclasses from battdat: https://github.com/ROVI-org/battery-data-toolkit/pull/135

``BatteryDataset`` now behaves like ``CellDataset`` (e.g., `data.cycle_stats` retrieves the "cycle_stats" table), but you build a dataset with pre-defined schema differently

I also moved us from conda to PyPI for building the environment, given that we no longer need conda-provided packages and I was getting intermittent build failures with conda

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -16,18 +16,12 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: conda-incubator/setup-miniconda@v3
+      - uses: actions/setup-python@v3
         with:
-            environment-file: dev/environment.yml
-            activate-environment: test
-            auto-activate-base: true
-            auto-update-conda: false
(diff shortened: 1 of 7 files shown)


.github/workflows/python-package.yml
```diff
@@ -16,18 +16,12 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: conda-incubator/setup-miniconda@v3
+      - uses: actions/setup-python@v3
```
victorventuri, line 19: Thank you! As someone who avoids conda as much as possible, seeing us also remove this sort of dependence is a relief 😄

victorventuri (approved):
Looks good, I take it the `extract_from_raw` method will come later?

PR #162, opened 2025-05-06
Add method to extract from raw data

### LINKED ISSUE #157: [FEATURE] Extractors require BatteryDataset objects
**Is your feature request related to a problem? Please describe.**
In order to use the extractors, the data needs to be passed as a `BatteryDataset` object. I vaguely recall having discussed this before, but I think this complicates things unnecessary, as most `BatteryDataset`s contain multiple cycles, but, in many cases, one wishes to only extract values for a single cycle (such as a low C-rate capacity check or an HPPC). Additionally, in some cases (the capacity one, for instance), assumes the `'cycle_stats'` have already been computed from the raw data, which is not always the case (for example, the synthetic data we recently generated). 

**Describe the solution you'd like**
Ideally, all off the extractors would not take in arguments that are not defined in `moirae` already. Perhaps a happy middle-ground would be they take in a `pandas.DataFrame` that has been formatted according to the `battdat` standard, but not an actual `BatteryDataset` object. 

**Describe alternatives you've considered**
Saving every cycle as a separate `BatteryDataset`, which is inefficient.

**Additional context**
I think that, in general, as much as we can, we should avoid requiring arguments and objects to be used in `moirae`'s main functionalities, sticking instead to things that are defined in the package (some are obviously unavoidable, like numpy arrays). We can create a separate part of the package dedicated to establishing the proper communication channels (like what we currently have with the interface to do online estimation directly from raw data). This should also help in cases where external packages are updated and/or modified outside our control (fewer places where things break)

### LINKED ISSUE #158: [BUG] OCV extractor assumes raw data contains derived features
**Describe the bug**
The `OCVExtractor` assumes the raw data contains the derived feature [`'cycle_capacity'`](https://github.com/ROVI-org/battery-data-toolkit/blob/main/battdat/schemas/column.py#L252), but that is not a required feature of a `Battery/CellDataset` object. 

**To Reproduce**
Import a `CellDataset` whose raw data only contains the [required columns ](https://github.com/ROVI-org/battery-data-toolkit/blob/main/battdat/schemas/column.py#L239), and try to extract OCV from it.

**Expected behavior**
The `OCVExtractor` should be able to work only with required columns of a battery dataset, that is, exclusively time, current, and voltage. (as well as other extracted quantities, such as capacity)

**Desktop (please complete the following information):**
 - OS: macOS Sequoia 15.4.1
 - Version: 0.0.1
 - Environment: Python 3.11.10

**Additional context**
Extractors should not require more than IV(t) (maybe temperature, depending on the model) and perhaps other extracted quantities. If needed, the `check_data` method must provide derived features.

WardLT:
Adds a method to the Extractors which requires a dataframe of current, voltage, time and cycle number rather than a full dataset (Closes #157 )

Also fixes the OCV extractor by having it compute the columns it requires (fixes #158 )

### DIFF
--- a/moirae/extractors/ecm.py
+++ b/moirae/extractors/ecm.py
@@ -5,7 +5,7 @@ import pandas as pd
 from sklearn.isotonic import IsotonicRegression
 from scipy.interpolate import LSQUnivariateSpline
 from scipy.optimize import differential_evolution
-from battdat.data import CellDataset
+from battdat.data import BatteryDataset
 from battdat.postprocess.integral import CapacityPerCycle, StateOfCharge
 from typing import Tuple
 
@@ -25,7 +25,7 @@ class MaxCapacityExtractor(BaseExtractor):
         2. Find the maximum capacity over all provided cycles
     """
 
-    def extract(self, data: CellDataset) -> MaxTheoreticalCapacity:
+    def extract(self, data: BatteryDataset) -> MaxTheoreticalCapacity:
         # Access or compute cycle-level capacity
         cycle_stats = data.tables.get('cycle_stats')
         if cycle_stats is None or 'capacity_charge' not in cycle_stats:
@@ -78,7 +78,7 @@ class OCVExtractor(BaseExtractor):
         self.soc_requirement = soc_requirement
         self.interpolation_style = interpolation_style
 
-    def check_data(self, data: CellDataset):
+    def check_data(self, data: BatteryDataset):
         # Check if there is a raw_data table
         if 'raw_data' not in data.tables:
             raise ValueError('`raw_data` table is required')
@@ -116,7 +116,7 @@ class OCVExtractor(BaseExtractor):
         model = IsotonicRegression(out_of_bounds='clip').fit(cycle['soc'], cycle['voltage'], sample_weight=w)
         return model.predict(self.soc_points)
 
-    def extract(self, dataset: CellDataset) -> OpenCircuitVoltage:
+    def extract(self, dataset: BatteryDataset) -> OpenCircuitVoltage:
         """Extract an estimate for the OCV of a cell
 
         Args:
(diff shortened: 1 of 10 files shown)


moirae/extractors/ecm.py
```diff
@@ -374,26 +377,26 @@ def interpolate_rc(self, data: BatteryDataset) -> np.ndarray:
...
+                    RCs[f'C{i_rc + 1}'].append(params_rc[2 * i_rc + 1])
 
         # Evaluate the smoothing spline for each element of each RC pair
         t = self.soc_points[1:-1]
 
         splines_eval = []
         for i_rc in range(self.n_rc):
-            Rspl = LSQUnivariateSpline(RCs['soc'], RCs[f'R{i_rc+1}'], t=t, k=1)
-            Cspl = LSQUnivariateSpline(RCs['soc'], RCs[f'C{i_rc+1}'], t=t, k=1)
+            Rspl = LSQUnivariateSpline(RCs['soc'], RCs[f'R{i_rc + 1}'], t=t, k=1)
```
WardLT, line 398: Sorry for this. My linter puts in whitespace, which flake doesn't check for :|

victorventuri: That's fine by me, personally I prefer the whitespace, but am not as familiar with all the linting rules around that

victorventuri (approved):
Nice, did not realize it would be so straightforward! The "casting" it as a `BatteryDataset` and working with that avoids editing much of the code, which is nice!

As a brief aside, we might want to open an issue to include Coulombic efficiency as an input to the extractors as well (by that I mean, for example, in the `__init__`), since it can modify the way we compute SOC. Not critical for now, as, in most cases, CE is basically equal to 100%, but for some experimental cells (looking at you, CAMP...) that assumption may not be valid. Or maybe the `battery-data-toolkit` `StateOfCharge` enhancer already does that?

WardLT:
The "casting" is something new, which I added in some other overhauls today.

I haven't touched Coulomb Efficiency since you noticed last year we had a pretty strong disagreement between the CE from battdat and those computed from the capacities determined by a tester. This was an old issue about it we should dust off at some point: https://github.com/ROVI-org/battery-data-toolkit/issues/58

victorventuri:
I don't quite remember that, but glad you do! Maybe we should be more explicit about that in that issue?

PR #165, opened 2025-05-09
Rc extract add stage and steps

victorventuri:
Minor changes so we can add the missing info from the new tagging systems in `batterry-data-toolkit`


moirae/extractors/ecm.py
```diff
@@ -7,7 +9,7 @@
 from scipy.optimize import differential_evolution
 from battdat.data import BatteryDataset
 from battdat.postprocess.integral import CapacityPerCycle, StateOfCharge
-from typing import Tuple
+from battdat.postprocess.tagging import AddSteps, AddState
```
WardLT, line 12: Thanks for moving the standard module to the top of the imports. Warms my pedantic heart!

WardLT (approved):

PR #166, opened 2025-05-09
Installs battdat from git

victorventuri:
This way, if we have new developments on the battery-data-toolkit that have not yet been published to PyPI, we can continue successfully using moirae.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ authors = [
     {name = "Noah Paulson", email = "npaulson@anl.gov"},
 ]
 dependencies = [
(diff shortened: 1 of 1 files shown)


WardLT:
Good call to put the installation in the pyproject.toml. We probably want users to get the latest version from GitHub until things settle

victorventuri:
That's more because I don't know where else to put it, given the test installs the dependencies from the `.toml`... I didn't think it would make sense to put it in the `test` section of that, since it would have already been installed, so idk if `pip` wouldn't just say "requirement met"

PR #167, opened 2025-05-14
Fixing a couple of things on RC extraction

victorventuri:
- Allows user to provide initial SOC
- Ensures consistent identification of rests

### DIFF
--- a/moirae/extractors/ecm.py
+++ b/moirae/extractors/ecm.py
@@ -429,7 +429,7 @@ class RCExtractor(BaseExtractor):
         cycle['soc'] = self.starting_soc + cycle['cycle_capacity'] / self.capacity  # Ensure data are [0, 1)
 
         if 'state' not in cycle.columns:
-            AddState().enhance(cycle)
+            AddState(rest_curr_threshold=self.max_rest_I).enhance(cycle)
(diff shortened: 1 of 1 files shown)


WardLT (approved):
This looks good to me, though makes me think of a new Issue: should we make an SOC extractor or at least break the logic for quick estimates of SOC into a separate function?

Merge if you think that's more than a quick fix, but feel free to make the change if it's trivial

victorventuri:
Noah and I briefly discussed this earlier, but I think there are parts of the logic here that, imo, really belong to the `battery-data-toolkit`. This PR is just a quick fix for me to get some other things going, but at some point, we should coordinate and see if some of the logic here should be ported over to the toolkit, and, if so, how and where to do it. 

I will wait for Noah's "ok" before merging.

npaulson:
It looks reasonable to me - are we making sure all the tests clear before merging?

victorventuri:
> are we making sure all the tests clear before merging?

The tests all pass on my side, so I think this might be related to a Coveralls issue @WardLT mentioned a while back. @WardLT is that the case here, and, if so, should we merge?

WardLT:
No, this one is probably due to an update from Thevenin. I'll fix that later

image

PR #169, opened 2025-05-23
Add utilities for reading estimated states from HDF5

### LINKED ISSUE #153: [FEATURE] Add utility functions for reading state estimates from HDF5
**Is your feature request related to a problem? Please describe.**
The HDF5 format in which we store state estimates requires significant HDF5 experience to use. 

Common operations to pull data from the files should be supported by functions.

**Describe the solution you'd like**

There are a few reading functions on the wish lists:
- [ ] Generate a stream of ASOH or transient states from the file
- [ ] Retrieve specific values (e.g., "mean of X", "covariance between X-Y", "mean and std of Z") from HDF5 into a Pandas dataframe

**Describe alternatives you've considered**
None

**Additional context**
@victorventuri gave me the wishlist

WardLT:
Fixes #153

### DIFF
--- a/moirae/interface/hdf5.py
+++ b/moirae/interface/hdf5.py
@@ -1,7 +1,8 @@
 """Tools for writing state estimates to HDF5 files"""
-from contextlib import AbstractContextManager
+from contextlib import AbstractContextManager, contextmanager
 from typing import Union, Optional, Literal, Dict, Any, Iterator
 from pathlib import Path
+import pickle as pkl
 import json
 
 from flatten_dict import flatten
@@ -9,8 +10,10 @@ from pydantic import BaseModel, Field, PrivateAttr
 import tables as tb
 import numpy as np
 
+from moirae import __version__
 from moirae.estimators.online import OnlineEstimator, MultivariateRandomDistribution
 from moirae.estimators.online.filters.distributions import MultivariateGaussian, DeltaDistribution
+from moirae.models.base import HealthVariable, GeneralContainer
 
 OutputType = Literal['full', 'mean_cov', 'mean_var', 'mean', 'none']
 
@@ -36,7 +39,6 @@ def _convert_state_to_numpy_dict(state: MultivariateRandomDistribution,
         raise ValueError('Mode cannot be none' if what == 'none' else f'Unrecognized what: {what}')
 
 
-# TODO (wardlt): Consider writing only every N timesteps
 class HDF5Writer(BaseModel, AbstractContextManager, arbitrary_types_allowed=True):
     """Write state estimation data to an HDF5 file incrementally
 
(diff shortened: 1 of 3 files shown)


tests/test_interface.py
```diff
@@ -236,6 +238,38 @@ def test_h5_read_what(simple_rint, tmpdir, what):
...
+    assert df.shape == (2, 4)  # Time, SOC, hyst, r0
+
+    df = read_state_estimates_to_df(h5_path, read_std=False, per_timestep=False)
+    assert len(df) == 1  # Only one cycle
+
+    # Test reading std and all cov
+    df = read_state_estimates_to_df(h5_path, read_std=True, read_cov=True)
+    assert len(df.columns) == 1 + 3 + 4 * 3 // 2  # Time, 3 means, upper tri of a 3x3 matrix
+
+    df = read_state_estimates_to_df(h5_path, read_std=False, read_cov=[('hyst', 'soc')])
```
victorventuri, line 264: I think it will pass, but can we add a test where we use `read_cov=[('soc', 'soc')]` and make sure it matches the squared value of the corresponding standard deviation? (sorta killing two birds with one stone, or, for the ornithologists in our midsts, kill two mosquitos with one slap)

victorventuri (changes requested):
Looks good, just think it would be nice to have a minor addition to the test, and left a couple of comments for possible future improvements/discussion

victorventuri:
Reading the outputs would also be nice!

moirae/interface/hdf5.py
```diff
@@ -265,6 +280,111 @@ def _unpack(row, typ):
...
+        data_path: Path to the HDF5 file or the group holding state estimates from an already-open file.
+        per_timestep: Whether to read per-timestep rather than per-cycle estimates
+    Yields:
+         The mean aSOH and battery state at each time step.
+    """
+
+    with _open_estimates_hdf5(data_path) as se_group:
+        # Get the output types for each
+        asoh_template: HealthVariable = se_group._v_attrs['initial_asoh_pkl']
+        tran_template: GeneralContainer = se_group._v_attrs['initial_transient_state_pkl']
```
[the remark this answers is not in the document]
WardLT: Good call. I'll store that as well.

moirae/interface/hdf5.py
```diff
@@ -265,6 +280,111 @@ def _unpack(row, typ):
...
+            for names in read_cov:
+                all_names.update(names)
+            unknown_names = all_names.difference(state_names)
+            if len(unknown_names) > 0:
+                raise ValueError(f'Covariance requests asks for terms not defined in names: {", ".join(unknown_names)}')
+
+            for st1, st2 in read_cov:
+                cov_to_read.append((
+                    state_names.index(st1),
+                    state_names.index(st2)
```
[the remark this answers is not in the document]
WardLT: Indeed, getting the order in the writing wrong would be _so_ annoying to debug. I'll add some tests to make sure it's correct now and stays correct in the future.

moirae/interface/hdf5.py
```diff
@@ -309,6 +308,84 @@ def read_asoh_transient_estimates(data_path: Union[str, Path, tb.Group],
...
+            ))
+    else:
+        raise ValueError(f'Unrecognized argument type for `read_cov`: {type(read_cov)}')
+
+    # Read them
+    for time, state, output in read_state_estimates(se_group, per_timestep):
+        data['time'].append(time)
+
+        # Decide whether to read state or outputs
+        row = state if read_state else output
```
victorventuri, line 365: But what if I want to read both states and the voltage predictions?

victorventuri (changes requested):
Nice work! Love the ability to save and re-load the estimator! Minor additional request is to be able to read both the output predictions and the hidden state at once if needed, rather than either-or

moirae/interface/hdf5.py
WardLT: I can make a function which opens the file once then calls the functions sequentially. Would that work?

I didn't combine the two into a single function because it would require ensuring that names are non-duplicated between transient/asoh/output (something we don't require now)

victorventuri: Not sure I fully understand it, but whichever solution you feel is best. I like the idea of being able to read any subset of the information stored, whether it is just a small part of the hidden states, or a mix of hidden and outputs, or everything. I imagine that, since we also store the estimator, some of the utility functions from that could help us re-build the relevant objects?

victorventuri (approved):

Issue #142, opened 2025-03-25
Add tutorial notebook for offline estimation


WardLT:
Describe how to build then use an offline estimator, describing the different kinds of loss functions available.

Expand what's already in here: https://github.com/ROVI-org/auto-soh/blob/main/notebooks/state-estimation-for-camp-cell.ipynb

WardLT:
Perhaps phrase the notebook around generating initial guesses for online estimation through a combination of extractors and offline estimation

PR #181, opened 2025-09-09
Offline rework

victorventuri:
This PR considerably changes the way offline estimation is performed. Now, we are more clear in specifying that every estimator that looks at past data to obtain estimates is considered an offline estimator. We have developed a framework for this kind of estimation, which works as follows:

1. First, the raw data is checked to make sure some conditions are met (for example, a full DoD charge and discharge step are present) and that the features necessary for parameter extraction are present (for example, pulses for extracting series resistance)
2. Once the data passes the check, it is used to extract raw parameter values, which are returned as a dictionary, detailing the value of the parameter, its unit, and, where appropriate, at which SOC level this value was obtained. In some cases, more information may also be returned.
3. These extracted values are passed on to an assembler, which creates an appropriate health variable. This two steps afford the user higher flexibility in defining how they wish to fit the extracted values
4. After all this is done, a full aSOH object is created. If the user desires, they can run further refinements over the aSOH, using a `scipy`-based minimizer over the mean squared error of the RPT voltage curve.

A few important points to note:

- The extractors naturally check the data, but it is possible to bypass this
- The extractors are meant to be sequential. For example, it is not possible to extract values of $R_0$ as a function of SOC if the capacity of the cell is unknown, since the SOC levels would be undetermined. 
- The overall framework assumes the user has some knowledge of the behavior of the extracted parameter. The assemblers need to be specified by the user. 

This PR also includes a notebook, `demonstrate_workflow.ipynb`, in which we
- Demonstrate how to generate data with `esynthm`, but do not require it, since the data is provided separately
- Demonstrate how to perform offline estimation with `moirae`'s tools
- Demonstrate how to use these offline-estimates to initialize an online estimator and run it over the data


victorventuri:
This also solves #142

